PR #19, opened 2024-02-06
update all models to use audio @ 30sec, and add aed support for nemo

nithinraok:
Three updates:
* updated all models to compute RTFX over multiple chunks of audio
* added aed support for nemo models
* moved from RTF to RTFX (higher the value, the more efficient model is)


nemo_asr/calc_rtf.py
```diff
@@ -135,65 +114,53 @@ def main():
...
-                        else:
-                            enc_out, enc_len, greedy_predictions = asr_model.forward(processed_signal=processed_example, processed_signal_length=processed_example_length)
-                            dec_out, dec_len = asr_model.decoding.ctc_decoder_predictions_tensor(
-                            enc_out, decoder_lengths=enc_len, return_hypotheses=False
-                            )
-                        torch.cuda.synchronize()
-                        end = time.time()
-                        if i >= warmup_batches:
-                            total_time += end - start
+        with torch.cuda.amp.autocast(dtype=torch.bfloat16):
```
titu1994, line 130: hmm, do we need to remove fp16 support?

titu1994 (commented):
LGTM, thanks !

nemo_asr/calc_rtf.py
nithinraok: We are not removing support, but only making sure to use ampere. This is done as we are only measuring RTF on A100.

nithinraok: @Vaibhavs10 to add hardware info on leaderboard space.

ctranslate2/calc_rtf.py
```diff
@@ -16,10 +16,10 @@
 ]
 
 n_batches = 3
-warmup_batches = 5
+warmup_batches = 3
 
-audio_file = "4469669.mp3"
-max_len = 600  # 10 minutes
+audio_file = "../data/sample_ami-es2015b.wav"
```
galv, line 21: Why change the audio file? What's wrong with the 4469669 file from EARNINGS 21?

nithinraok: not ccby4, so we don;t have permissions to redistribute. AMI is CC-by-4

galv:
IMO, instead of moving everything to batch size 1 inference, we should run NeMo's models at larger batch sizes. It is indeed silly that NeMo models currently run at batch size 1 while Whisper models run at batch size 24, but nerfing the Whisper models doesn't seem right to me for showing off what these models can really do. Why not also chunk the input into 30 second chunks for NeMo the way that Whisper is?

In the future, we should come up with a more precise benchmark the way that MLPerf does it. I did the MLPerf Inference benchmark for RNN-T a few years ago and have suggestions.

galv:
Another problem is that running a single example multiple times can have more cache hits than you would say in production in the embedding layers. I'm not sure if that is actually happening here. But the easiest way to prevent these kinds of weird issues is to run on several different inputs (sort of like what was happening before, but ideally at a larger time scale than a 10 minute recording).

ctranslate2/calc_rtf.py
galv: Ah, good call there.

Slyne:
Why not just test the RTF on the test datasets (Librispeech, AMI, Gigaspeech, .....)?
There are 2 good reasons to do so:
1. These datasets have audio files of various length. So this can more comprehensively evaulate the performance.
2. It will cause less misunderstanding when users see RTF column with the other dataset columns together on the Open ASR leaderboard.

And is it possible to use RTFX - the inverse of RTF? It will be easier for people to interpret integers. For example, `RTF=0.01`, then `RTFX=100`.

sanchit-gandhi:
I would be in favour of measuring the RTF(X) as @Slyne has described (i.e. by averaging over all the short-form test sets) => this then gives an RTF(X) value that is consistent with the short-form WERs we show on the leaderboard. It's also then easier for the user to reproduce, since only a single script has to be run.

cc @Vaibhavs10

Vaibhavs10:
Thanks all for the help! Sounds good to me. 
Let's open a PR to do this in transformers too :)

nithinraok:
@Slyne @galv @sanchit-gandhi @Vaibhavs10 

1. I updated the PR to calculate RTFX over the average of multiple chunks (30 sec) of longer audio (10 min). This will ensure that the model's output of RTFX is less dependent on the type of speech spoken in the audio sample.
2. @Slyne, calculating over a benchmark might not be a good idea, as it takes a lot of time to process any benchmark, and averaging over multiple times over that benchmark for RTFX will take even more time.
 
@Vaibhavs10 I have updated code for transformers as well, make sure it is computed correctly as well. 

Next step is to update RTFX and Hardware information on the leaderboard.

sanchit-gandhi:
Hey @nithinraok - in #20 I've updated the Transformers short-form evaluation code (`run_eval.py`) to compute the WER **and** RTF over the short-form datasets. It gives an output along the lines of:
```
Filtering models by id: openai/whisper-tiny.en
********************************************************************************
Results per dataset:
********************************************************************************
openai/whisper-tiny.en | open-asr-leaderboard-datasets-test-only_common_voice_test: WER = 19.64 %, RTFX = 108.91
openai/whisper-tiny.en | open-asr-leaderboard-datasets-test-only_librispeech_test.clea: WER = 4.76 %, RTFX = 105.03
openai/whisper-tiny.en | open-asr-leaderboard-datasets-test-only_librispeech_test.other: WER = 7.78 %, RTFX = 89.11

********************************************************************************
Composite Results:
********************************************************************************
openai/whisper-tiny.en: WER = 10.73 %
openai/whisper-tiny.en: RTFX = 101.02
********************************************************************************
```

=> as discussed, computing the WER and RTF over the same detests is a more consistent approach to our metrics (rather than doing short-form WER, but long-form RTF)

Feel free to take a look at the PR - I'd be happy to help propagate the changes forward to NeMo as well!

nithinraok:
@sanchit-gandhi I mentioned some issues for measuring RTF benchmark on datasets, hence proposal is to iterate over chunks multiple times (to account for silence/amount of speech in audio).

sanchit-gandhi:
Hey @nithinraok! By averaging the RTFX over **multiple datasets** that cover multiple speaking styles and domains, surely we provide a more representative RTFX value than a single 10 min audio file? If we compute the RTFX on the short-form test sets, then we average the RTFX over **thousands** of audio samples -> this will surely give us a more representative metric for RTF, one that takes into account silence, speaking style, accents etc.

Note that computing the RTF adds no overhead to the benchmark time, since there's no need to do multiple loops of each dataset. Each dataset is thousands of samples, we can be pretty certain that the RTF averages out to a representative value over one pass of each dataset.

nemo_asr/calc_rtf.py
```diff
@@ -135,70 +96,61 @@ def main():
...
-        with profiling_context:
-            with precision_context:
-                with torch.no_grad():
-                    for i in tqdm(range(nbatches + warmup_batches)):
-
+        with torch.cuda.amp.autocast(dtype=torch.bfloat16):
+            with torch.no_grad():
+                for i in tqdm(range(nbatches + warmup_batches), desc='Calculating RTF', unit='batch'):
+                    for idx in range(0, total_audio_len, chunk_len): # process audio in chunks
+                        chunk_singal = input_example[:, idx*SAMPLING_RATE:(idx+chunk_len)*SAMPLING_RATE]
```
galv, line 115: singal -> signal

nemo_asr/calc_rtf.py
```diff
@@ -135,70 +96,61 @@ def main():
...
+                        chunk_signal_length = torch.tensor(chunk_len * SAMPLING_RATE).to(DEVICE).repeat(batch_size)
```
galv, line 116: You're putting a synchronous operation into a hot path of your code here FYI. Can be bad for performance.

galv:
> Note that computing the RTF adds no overhead to the benchmark time, since there's no need to do multiple loops of each dataset. Each dataset is thousands of samples, we can be pretty certain that the RTF averages out to a representative value over one pass of each dataset.

There are a few exceptions I can think of: You may want some "warmup", so doing two iterations over a dataset rather than once is the easiest way to implement that. Secondly, when you run super duper fast (think 4000 RTFx) but your dataset is only 5.4 hours (like librispeech test other), averaging over multiple iterations can be helpful, since each run will take only 5 seconds. I encountered that exact scenario in https://arxiv.org/pdf/2311.04996.pdf 
image

In order to get stable measurements, I had to iterate over my datasets multiple times.

The easy fix (which I believe MLPerf does) is to require a model to do inference for at least some threshold number of minutes, repeating over the dataset as required.

galv:
I guess I just have one concern with changing the audio file @nithinraok . Sure, that original file is not CC-BY, so we cannot include it in this repo. But I don't think that is a good reason to change the underlying data on which RTF is measured. We can change the data next time, for the short-form and long-form RTF evaluations, as discussed. But whey not keep the underlying data the same for now? Otherwise, all models other than whisper and your Canary model will need to be reevaluated on this new AMI data file.

nemo_asr/calc_rtf.py
nithinraok: Do you recommend through iterator? However measurement of duration is outside it.

nithinraok:
I rechecked the license, on HF datasets I could see [earnings22](https://huggingface.co/datasets/anton-l/earnings22) as CC-by-SA-4.0, which means we can redistribute and modify. No license on their github page, where data was originally posted: https://github.com/revdotcom/speech-datasets

titu1994:
@sanchit-gandhi I want to avoid combining the task of wer calculation and the task of RTF calculation into single script or process. It's might be more favorable for models like whisper which have fixed size input and output and therefore can benefit from specialized code while Nemo and other libraries have to deal with padding and chunking effects. 

Furthermore, the evaluation of a model in a bunch of datasets for rtf doesn't make sense - different models behave in different ways due to architecture. For example, Nemo parakeet can concatenate entire test set of LS and do single forward pass over it and get an incredible RTF - that's still incorrect resulfs though cause of an architecture benefit. 

Another thing is batch size for inference, Nemo can tolerate super high batch sizes but may be slightly slower or faster at BS 1 - by combining the two you've put a limiter on the actual perf of the model. Speech brain for example can't go above BS 4 - when Nemo can do BS 32 with ease. 

Let's please keep RTF and WER calculation as two different steps.

nemo_asr/calc_rtf.py
galv: I would say don't worry about it for now. I just wanted to alert you to it as an "FYI" sort of thing.

galv:
This PR at this point accomplishes two things:

1. It allows nemo models to transcribe "chunk_len" seconds at a time instead of transcribing the entire 600 second audio. Running on a chunk which is smaller than the entire audio size is important in particular for new NeMo AED models, becuase of the quadratic nature of attention (as I understand it)
2. It reverts the current behavior in the HF transformers benchmark code of running at batch size 24 (notably, there are 24 30 second chunks in the 600 second audio) to batch size 1. https://github.com/huggingface/open_asr_leaderboard/pull/19/files#diff-2b020a022ac15c236e5d6e55ccb2e85d4f34a7ffde1f816a2db34cca0b370c90R57 While this is not ideal for transformers's Whisper implementation because it makes the performance look worse compared to before, I sincerely hope that you can agree that this is "right" by the current RTF measurement standards. The discussion about short form RTF measurement can continue to take place in #20 .

So I would ask that this PR be accepted in its current state.

titu1994:
Agreed on the fact that running Whisper at chunk batch size 1 does look worse than before, but it is a fairer comparison between models (and something most ASR models, HF and NeMo not withstanding, can do inference in this mode, making it future proof).

I also agree that we need to re-add RTF for short form datasets, but this PR is not the place for it. The link above can carry out discussion, and we can add a new script (or same script with mode option or something) that satisfies all parties requirements of a valid stable benchmark.

Again, this should be for RTF. I do not believe it is wise to combine RTF and WER calculation in the same script due to multiple reasons discussed above.

titu1994:
> I don't really see how evaluating RTF over a single 10 minute audio file solves this? To me, averaging RTF over thousands of audio samples from 10 datasets gives a more representative figure for the latency of the model.

Ok, we can calculate RTF over one dataset, that too is sufficient no? And I mean only calculate RTF on a standard dataset like Librispeech test other, not the wer at the same time. Higher batch sizes can cause transcription effects in TDTs due to its batching algo. Due to BN usage in many Conformer models, wer is affected with more batch size and more padding. If we're doing dataset level RTF calculation, then let's just stick to RTF calculation. 

> Happy to set the maximum possible batch size for the RTFX calculation as discussed https://github.com/huggingface/open_asr_leaderboard/pull/20#discussion_r1482174219. Otherwise, how do you propose setting the batch size for the RTF calculation in this PR? It faces the same issues you've described above

My question is that while this is great from an MLPerf point of view to see what's the fastest model, how is it fair to compare completely different architectures like this? Daniels approach is fair iff we follow MLPerf to a T - same model, trained on same data for same steps and only hardware changes, not decoding algo. At that point, yes, you are fairly testing your accelerates max perf on the same benchmark. 

This is not thr case for this benchmark. The decoding algos are different, the architectures are different, sizes are different. The only constant we can perform are the eval datasets and batch size. This is not Daniels use case, this is more diverse comparison. Even in the paper, he keeps every single thing constant, changing only the WFST deciding algo. 

So for a fair comparison, a large batch size should be picked, same dataset should be picked, and a simple standard decoding algorithm should be picked that can be done by any generic ASR model. For simple standard decoding algorithm, sequential chunks is a reasonable processing technique - which doesn't depend on wer correctness just the throughput of how many chunks you can process in finite data. Batched chunks are a special case that only few models like whisper can do with advanced algos from HF finely tuned for the task - rnnt, tdt, naive AED from Nemo and speech brain can't implement that easily or correctly for wer so again we should not calculate wer during RTF calculation. 

> Since it's short-form evaluation, there's no chunking, and batch size simply refers to file batch size. To me, this is the most fair comparison across implementations: we take out the complexity of chunking altogether, and just evaluate on the audio segments as they are in the short-form datasets. Everything is a level playing-field then: the model is only required to evaluate batches of <30s audio samples.

On this, we agree. This is the fairest comparison of the simplest and most common use case, simple inference on a single file or a fixed batch of files

> As you say, the user is then free to write any chunking code on-top of the provided {Transformers, NeMo, ESPNet, ...} implementations, but we don't want to count for this in our benchmark. Let's just give a representation of how fast the model is in each library by computing the RTF for the short-form audio samples

This can be yet another benchmark - maybe the long form variant. A general algo as I meant above can be simply taking 10 different long audio files of different lengths, chunking them all to 30 sec, and sequentially / parallel processing them - however all models just use the same batch size and the same hard coded decoding loop - no using internal optimized library codepath that HF or Nemo script etc. We write an explicit batched chunk loop and execute the models forward pass to decoder text for each model to determine the rtf for this scenario. This kernels data and algo constant but allows models to differ on raw perf. ..

> Overall I do see where you're coming from, but I don't agree with the notion of having a WER that is calculated using one configuration, and an RTF that is calculated using another. In my opinion, the RTF should be reflective of the configuration used to achieve the transcription decoding, i.e. the one used to compute the WER. If we have a WER computed using one configuration, and an RTF computed using another configuration optimised for throughput, it's not really indicative to the user what performance they'd actually get when using this model downstream.

There is always a tradeoff between more perf and accuracy - your HF implementation itself said there is about a 1% wer degradation between chunked batches vs sequential chunks. That's why I want to disentangle wer and RTF. Wer is the concrete representation of a models accuracy under a constrained batch size (sample level, not chunk level batching) for all models of whatever size, but on same dataset. The algo is also general - either Autoregressive decoding with rnnt tdt or aed or nar with CTC, doesn't matter, we only compare text at the end to get wer. 

RTF is the result of optimization that increases wer, it's the maximal efficient inference speed but degrades wer for that sake so it doesn't make sense to conflate the two. Even in MLPerf, the only way they can do both wer and speed comparison at the same time is because they have standardized reference model implementation - everyone follows the same model arch and implemented the decoding same way. All things are kept same - only hardware is different. We do not have the same criterion as MLPerf - all models are different, decoding is different. So conflating wer and speed is not correct

titu1994:
Forgive the spelling mistakes and typos, I'm typing at 3.30 AM PST 😓

sanchit-gandhi:
Thanks for taking the time to reply in such detail @titu1994!

> Due to BN usage in many Conformer models, wer is affected with more batch size and more padding. 

Interesting - do you know (approx.) how much this penalty is? We also see a degradation for Wav2Vec2 with dynamic padding, but it's on the order of ~0.1% (i.e. largely insignificant). We could explore @galv's suggestion of sorting the short-form samples by length, and then transcribing audios of similar length in each batch. This would reduce the penalty for all models that use padding/BN. WDYT?

> So for a fair comparison, a large batch size should be picked, same dataset should be picked, and a simple standard decoding algorithm should be picked that can be done by any generic ASR model.

Great - I think we're both in agreement on fixing:
1. **Dataset(s):** either LibriSpeech ASR, or all 10 of the OpenASR short-form test sets
2. **Decoding algorithm:** no chunking (this is fine since we're doing short-form eval)
3. **Batch size:** pick a large batch size and keep consistent for all models

> This can be yet another benchmark - maybe the long form variant.

Agreed. Let's remove all ambiguity in the current benchmark and compute the WER and RTF for the same set-up: short-form audio samples in a batch with no chunking.

> RTF is the result of optimization that increases wer, it's the maximal efficient inference speed but degrades wer for that sake so it doesn't make sense to conflate the two

Here's a counter argument: typically with Whisper, we generate 128 tokens per utterance. In the RTF calculation, I could limit the maximum number of generated tokens to just 64. While this would decrease the latency by 2x (generating half as many tokens), it would double my WER. However, we wouldn't see this difference in our results, since we're just computing the RTF. Is this not the same as the penalty you see when increasing the batch size, just on a higher scale? The model could be generating useless predictions in the RTF calculation, but still be getting a ridiculously good RTF score? Should we not be giving a representative score for both?

titu1994:
Anyway let's pause the hypothetical scenario the moment. If you insist on having wer calculation as a check, what we can instead do is this 

1) calculate wer as a criterion for passing the test. No point putting a threshold.
2) pick exactly one modest sized dataset for eval per language. For English it can be LS test other v
3) Eval both rtf and wer on this benchmark with the calc RTF code. 
4) Asset that the wer from RTF calculation is+- 5% relative with offline calculated WER on same dataset using run_xyz.py script 
5) Don't release the wer of RTF calculation, instead release the wer of offline calculation with run script in the leaderboard.

That's it. That satisfies your requirement to calculate RTF, prevents gaming the benchmark, and keeps RTF calculation to a reasonable cost and complexity over a single dataset

sanchit-gandhi:
> It increases with increase in batch size. BS 32 is about 0.1% off for ctc, 0.2% off with RNNT, and around 0.2 to 0.3% off with tdt.

Great - thanks for the numbers! What's the recommendation for practitioners running the model with high batch sizes? Do you advise they sort their data by length to minimise padding? Or are these "silent" regressions that are an artefact of the architecture? If it's the latter, IMO it's fair if the leaderboard reflects them!

>  Different models hand different speeds, to calculate RTF should not take 10-24 hours

We're already waiting 10-24 hours for the WER. There would be no additional benchmark time if we computed WER and RTF jointly over all datasets. If we want an indication of RTF quickly, we can just iterate using one dataset, and tune our batch size quickly based on that. Once we have the batch size fixed, we can progress to evaluating over all datasets to get our final number. 

To me, this is the most fair way of reporting these metrics. There's no arbitrary +-5% threshold set by us, it's the pair of (WER, RTF) achieved in the evaluation. It's also easier for new library additions, since only one new script has to be added.

> It is better to stick to a realistic situation where benchmark code is inspected in public and then run by each party 

I agree that that the RTF/WER should be standardise to use the same inputs and no hidden-tricks. One key part of the leaderboard design is that contributing new libraries is easy, since you can drop in your inference API into the template code. Therefore, I wouldn't go so far as restricting the use of higher-level APIs, as this might hinder new library additions, but I would make sure that at least 2 people working on different review the benchmark code for each library to ensure "fair" arguments are set. I also want to reiterate that by using short-form datasets, we take away all the ambiguity for chunking/long-form strategies, so it's just a case of making sure that the HPs are set appropriately.

sanchit-gandhi:
For interest, I ran the joint RTF/WER benchmark as proposed in #20 for Distil-Whisper. The RTFs do vary quite substantially over datasets, since it's a function of the audio data as well as the model decoding. Here's the full list of RTFs for Distil-Whisper:

```
Filtering models by id: distil-whisper/distil-large-v2
********************************************************************************
Results per dataset:
********************************************************************************
distil/whisper-distil-large-v2 | ami_test: WER = 14.67 %, RTFX = 88.95
distil/whisper-distil-large-v2 | common_voice_test: WER = 12.61 %, RTFX = 196.84
distil/whisper-distil-large-v2 | earnings22_test: WER = 12.14 %, RTFX = 212.39
distil/whisper-distil-large-v2 | gigaspeech_test: WER = 10.31 %, RTFX = 195.72
distil/whisper-distil-large-v2 | librispeech_test.clea: WER = 2.94 %, RTFX = 222.33
distil/whisper-distil-large-v2 | librispeech_test.other: WER = 6.40 %, RTFX = 196.10
distil/whisper-distil-large-v2 | spgispeech_test: WER = 3.28 %, RTFX = 287.59
distil/whisper-distil-large-v2 | tedlium_test: WER = 4.30 %, RTFX = 241.87
distil/whisper-distil-large-v2 | voxpopuli_test: WER = 8.22 %, RTFX = 278.94

********************************************************************************
Composite Results:
********************************************************************************
distil/whisper-distil-large-v2: WER = 8.32 %
distil/whisper-distil-large-v2: RTFX = 213.41
********************************************************************************
```

=> there's a 70% difference between the highest and lowest RTFX. By averaging over these datasets, we'll give a more representative number of how the model performs globally over different audio conditions, speaking styles, accents, etc

sanchit-gandhi:
> I rechecked the license, on HF datasets I could see [earnings22](https://huggingface.co/datasets/anton-l/earnings22) as CC-by-SA-4.0... No license on their github page, where data was originally posted: https://github.com/revdotcom/speech-datasets

Thanks to @qmac for clarifying that the licence for both Earnings-21 and 22 are indeed CC-BY-SA 4.0, as denoted in the subfolder for each dataset on the [revdotcom/speech-datasets](https://github.com/revdotcom/speech-datasets) GH repo

E.g. for Earnings-22: https://github.com/revdotcom/speech-datasets/tree/main/earnings22#earnings-22

PR #25, opened 2024-05-15
Added PNC (punctuation and capitalization) flag for NeMo ASR model inference

KunalDhawan:
**What this PR does:** Adds a flag for specifying PnC mode (output with punctuation and capitalization) for NeMo ASR model inference.

**Why is it required:** The default mode for NeMo Canary inference is output with Punctuation and Capitalization, whereas the leaderboard numbers are reported with pnc mode turned off. This causes a mismatch in the reported numbers and the numbers obtained from current eval script. 
For e.g., for Librispeech-test-clean, the current version of the eval script gives a WER of 1.56% which is inconsistent with the reported numbers. With the provided fix (setting default Canary pnc inference to false), we get WER of 1.49%, as expected.

### DIFF
--- a/nemo_asr/run_eval.py
+++ b/nemo_asr/run_eval.py
@@ -54,7 +54,7 @@ def pack_results(results: list, buffer, transcriptions):
     return results
 
 
-def buffer_audio_and_transcribe(model: ASRModel, dataset, batch_size: int, cache_prefix: str, verbose: bool = True):
+def buffer_audio_and_transcribe(model: ASRModel, dataset, batch_size: int, pnc:bool, cache_prefix: str, verbose: bool = True):
     buffer = []
     results = []
     for sample in tqdm(dataset_iterator(dataset), desc='Evaluating: Sample id', unit='', disable=not verbose):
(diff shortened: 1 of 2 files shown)


nemo_asr/run_eval.py
```diff
@@ -54,15 +54,19 @@ def pack_results(results: list, buffer, transcriptions):
...
     buffer = []
     results = []
     for sample in tqdm(dataset_iterator(dataset), desc='Evaluating: Sample id', unit='', disable=not verbose):
         buffer.append(sample)
 
         if len(buffer) == batch_size:
             filepaths = write_audio(buffer, cache_prefix)
-            transcriptions = model.transcribe(filepaths, batch_size=batch_size, verbose=False)
+
+            if pnc is not None:
```
nithinraok, line 66: Minor for cleaner look: Instead of if else for each argument update PnC as true if PnC is None. And pass to func args

KunalDhawan: Thanks for the feedback! Chose this design to support FastConformer-CTC and FastConformer-RNNT (and other future models) that don't support pnc argument in their transcribe function. This allows us to use the same buffer_audio_and_transcribe() function for all NeMo ASR models.

PR #27, opened 2024-07-04
[transformers] enable static k/v

sanchit-gandhi:
@Vaibhavs10 this PR shows the changes in `run_eval.py` that are required to enable torch compile for Transformer models (CTC + Seq2Seq). Note that it's opened on top of #20, meaning we don't actually want to merge this PR, but wait till #20 is merged and then create a new PR with the changes in `run_eval.py` on top of `main`.


transformers/run_eval.py
```diff
@@ -1,49 +1,152 @@
 import argparse
```
sanchit-gandhi, line 1: This is the only file of interest for this draft PR. It should form a new PR on top of `main` once #20 is merged

transformers/run_whisper.sh
```diff
@@ -13,7 +13,7 @@ do
 
     python run_eval.py \
```
sanchit-gandhi, line 14: Every script then requires us adding one arg to enable compilation:
```sh
--torch_compile \
```

sanchit-gandhi:
Closing in favour of #32

PR #29, opened 2024-07-31
Use faster-whisper model names instead of hardcoded HF names

jordimas:
Use model names instead of hardcoded HF names.

Nowadays for example faster-whisper does not use guillaumekln/faster-whisper-tiny.en uses Systran/faster-whisper-tiny.en models since guillaumekln was associated to a personal account no longer active in the project.

Also adding large-v3 model.


sanchit-gandhi (approved):
Thanks @jordimas!

PR #30, opened 2024-07-31
Switch to hf-audio/esb-datasets-test-only-sorted dataset

jordimas:
Switch to audio/esb-datasets-test-only-sorted dataset which is the one that I understands is used for the rest of the implementations for benchmarking


sanchit-gandhi (approved):
Thanks @jordimas! Feel free to merge after removing common voice from the list of benchmark datasets.

The dataset we're using for benchmarking is simply the sorted version of the one we had before. This ensures an OOM happens on the first decoding step (if there is to be one), helping when setting the batch size.

jordimas:
Changes done

It says "Only those with to this repository can merge pull requests."

sanchit-gandhi:
Thanks @jordimas!

PR #37, opened 2024-08-12
[transformers] update bash scripts

sanchit-gandhi:
1. Run VoxPopuli first, since it has the longest audio sample overall, so an OOM happens on the first evaluation step
2. Split Wav2Vec2 and Wav2Vec2-Conformer, since they are different architectures and thus should have different bash scripts
3. Update Whisper to include `large-v3` and all distil-whisper models


Vaibhavs10 (approved):
Thanks for the updates! 🤗

PR #41, opened 2024-10-14
add pandas to requirements/requirements.txt

michael-james-smith:
refactor normalizer/eval_utils.py to use pandas
and output results to a csv file


michael-james-smith:
@scottlimmer, would you mind taking a look at this and let me know what you think of the changes.

PR #43, opened 2024-11-02
Add UsefulSensors Moonshine benchmark

njeffrie:
Create a new evaluation runner and shell script to evaluate Useful Sensors `Moonshine` models. For reference see the [Moonshine repo](https://github.com/usefulsensors/moonshine).

Script is based on transformers `run_eval.py` with some changes:
- Add `trust_remote_code=True` since our models are not yet part of the `transformers` repo.
- Remove batching support until we add batch decoding
- Pad inputs to 1024 audio samples since `Moonshine` uses a variable length encoder which requires this minimum length.
- Use tokenizer directly rather than processor, since `Moonshine` uses a trainable feature extractor that is part of the model rather than a mel spectrogram.

Until we add batch decoding support, RTFx values will be artificially low, but accuracy numbers will be representative. All WER values match those published in the [Moonshine paper](https://arxiv.org/pdf/2410.15608) when `eval_moonshine.sh` is run on an H100 instance.

### DIFF
--- a/moonshine/run_eval.py
+++ b/moonshine/run_eval.py
@@ -1,0 +1,192 @@
+import argparse
+import os
+import torch
+from transformers import AutoConfig, AutoModelForSpeechSeq2Seq, AutoProcessor, PreTrainedTokenizerFast
+import evaluate
+from normalizer import data_utils
+import time
+from tqdm import tqdm
+import numpy as np
+
+wer_metric = evaluate.load("wer")
+torch.set_float32_matmul_precision('high')
+
+def main(args):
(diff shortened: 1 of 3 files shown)


njeffrie:
@sanchit-gandhi or @Vaibhavs10 could you please take a look and let me know if there are any changes necessary to get this into the next leaderboard release?

Thanks!

Vaibhavs10 (commented):
Hi @njeffrie - What's the progress on Batch Decoding? We'll make a new release in the coming two weeks.

PR #49, opened 2025-02-06
Adding largescale ASR model for speechbrain

TParcollet:
Hello there,

This is adding the scripts and modifications to run the evaluation of https://huggingface.co/speechbrain/asr-conformer-largescaleasr . I don't have an A100 available right now ... AND some of the datasets should not be used on company clusters due to license. Any chance that you could run the evaluation?

Thanks!

### DIFF
--- a/speechbrain/run_conformer.sh
+++ b/speechbrain/run_conformer.sh
@@ -3,7 +3,7 @@
 export PYTHONPATH="..":$PYTHONPATH
 
 SOURCE="speechbrain/asr-conformer-largescaleasr"
-BATCH_SIZE=8
+BATCH_SIZE=32
 DEVICE_ID=0
 
 # Run with CTC+Attn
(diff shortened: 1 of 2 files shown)


speechbrain/run_conformer.sh
```diff
@@ -2,27 +2,32 @@
...
 
+# Run with CTC+Attn
 python run_eval.py \
   --source=$SOURCE \
   --speechbrain_pretrained_class_name="EncoderDecoderASR" \
   --dataset_path="hf-audio/esb-datasets-test-only-sorted" \
   --dataset="librispeech" \
   --split="test.clean" \
-  --device=0 \
+  --device=7 \
```
nithinraok, line 14: revert this as 0

speechbrain/run_conformer.sh
```diff
@@ -2,27 +2,32 @@
...
   --batch_size=4 \
-  --max_eval_samples=-1
+  --max_eval_samples=-1 \
+  --beam_size=40 \
```
nithinraok, line 17: Isn;t this too high of a beam size :D (can we do with 5, curious as 40 would take lot of time to eval)

speechbrain/run_conformer.sh
```diff
@@ -2,27 +2,32 @@
...
 
+# Run with Attn only
 python run_eval.py \
   --source=$SOURCE \
   --speechbrain_pretrained_class_name="EncoderDecoderASR" \
   --dataset_path="hf-audio/esb-datasets-test-only-sorted" \
   --dataset="librispeech" \
-  --split="test.other" \
-  --device=0 \
+  --split="test.clean" \
```
nithinraok, line 25: Could you include all datasets (see example here: https://github.com/huggingface/open_asr_leaderboard/blob/main/nemo_asr/run_fast_conformer_rnnt.sh)

nithinraok (changes requested):

speechbrain/run_conformer.sh
nithinraok, line 1: since you are using same model but different settings, I don;t think this would play well when HF outputs results as the files are saved based on model_name and eval data. Pls run on your machine to verify this. Otherwise rest all LGTM

nithinraok (approved):

Vaibhavs10 (approved):

speechbrain/run_conformer.sh
TParcollet: To be honest, i've played a bit with a local a100, and the decoding seems to slow compared to what I would have expected (what we see within SB). I'll investigate a bit.

TParcollet: Yes please don't run any test - there is something fishy. I'll investigate.

TParcollet:
all fixed @Vaibhavs10

PR #54, opened 2025-03-10
Fix attention mask

o-alexandre-felipe:
Fix the WARNING
> The attention mask is not set and cannot be inferred from input because pad token is same as eos token. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.

The fix was very simple, passing `return_attention_masks=True` also when `model.can_generate()`

I also wrapped `MODEL_IDs` in the bash scripts to make it readable without scrolling horizontally.


Vaibhavs10:
Closing this since this is fixed in main.

PR #56, opened 2025-03-12
adds suppport of beam search with configurable num_beams for Phi-4-Mu…

freewym:
…ltimodal. There was an assertion error for microsoft/Phi-4-multimodal-instruct when num_beams > 1. We fixed that error [here](https://huggingface.co/microsoft/Phi-4-multimodal-instruct/commit/18812f44eec22f4347a85536d97059356bd5ec2f) . This PR was to show how to set num_beams for beam search.

### DIFF
--- a/phi/run_eval.py
+++ b/phi/run_eval.py
@@ -57,7 +57,7 @@ def main(args):
 
     prompt = f"{user}<|audio_1|>{args.user_prompt}{prompt_suffix}{assistant}"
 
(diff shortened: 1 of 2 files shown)


freewym:
@Vaibhavs10 We fixed an assertion error in Phi-4-multimodal model, and would like to also update the corresponding script here.

Deep-unlearning:
LGTM! Thanks for the update and for fixing this assertion error

PR #99, opened 2025-09-22
increases batch size for Phi-4-Multimodal for better RTFx and WER

freewym:


freewym:
WER = 6.02 %  RTFx = 151.10

PR #142, opened 2026-04-13
Switch to HF jobs + normalizer improvements

ebezzam:
## Description

- Standardizing usage of Transformers models for English and multilingual: Single `run_eval.py` file for all models supported by Transformers (rather than separate folders).
- Use HF jobs for better reproducibility + upgrading hardware to h200.
- Save file name for easier analysis (applicable for all models).
- Better timing with `torch.cuda.Event`.
- Improved normalizing mappings and merging compounds with [kaldialign](https://github.com/pzelasko/kaldialign#compound-word-matching).


transformers/run_eval.py
```diff
@@ -2,61 +2,162 @@
...
+            texts=[text] * minibatch_size
+        else:
+            texts = None
 
         # START TIMING
-        start_time = time.time()
+        torch.cuda.synchronize(device=args.device)
+        start_event = torch.cuda.Event(enable_timing=True)
+        end_event = torch.cuda.Event(enable_timing=True)
+        start_event.record()
```
ebezzam, line 168: Update timing approach to CUDA methods. Shouldn't make a difference from before, as we anyway had to wait for decoding, but still more precise.

transformers/run_eval.py
```diff
@@ -2,61 +2,162 @@
...
     def benchmark(batch, min_new_tokens=None):
         # Load audio inputs
         audios = [audio["array"] for audio in batch["audio"]]
         minibatch_size = len(audios)
-        
-        # Compute audio length in seconds (16kHz sampling rate)
-        batch["audio_length_s"] = [len(audio) / 16_000 for audio in audios]
+        sampling_rate = batch["audio"][0]["sampling_rate"]
+        batch["audio_length_s"] = [len(audio) / sampling_rate for audio in audios]
+        batch["audio_filepath"] = data_utils.extract_audio_filepaths_from_batch(batch, minibatch_size)
```
ebezzam, line 158: New method to extract file name for future analysis!

transformers/run_eval_ml.py
ebezzam, line 1: update to handle Voxtral, Voxtral Realtime, and CTC models

ebezzam (commented):
@Deep-unlearning small review if it helps

normalizer/data_utils.py
```diff
@@ -87,3 +88,53 @@ def prepare_data(dataset, sampling_rate=16000):
     return dataset
 
 
+AUDIO_FILEPATH_METADATA_KEYS = [
+    "id",           # Main: https://huggingface.co/datasets/hf-audio/open-asr-leaderboard
+    "file_name",    # Multilingual: https://huggingface.co/datasets/nithinraok/asr-leaderboard-datasets
+    "original_filename", # Private
+    "file_name",    # Private
```
Deep-unlearning: duplicate on the key

Deep-unlearning, line 97: duplicate on the key

transformers/run_eval.py
```diff
@@ -254,4 +393,8 @@ def benchmark(batch, min_new_tokens=None):
     args = parser.parse_args()
     parser.set_defaults(streaming=False)
```
Deep-unlearning: `parser.set_defaults(...)` after `parse_args()` is a no-op — `args` is already materialized, so this line never affects `args.streaming`.

Combined with `action="store_false"` on `--no-streaming` (lines 347-352 above), which has an implicit default of `True`, this means `args.streaming` is always `True` unless `--no-streaming` is passed — and none of the shell scripts pass it. Every eval is running in streaming mode, with no way to opt out from a script.

For a benchmark we want pinned data (no network jitter in RTFx, reproducible timings). Recommend flipping to an opt-in `--streaming` flag with non-streaming as the default. The flag rename at lines 347-352 needs to be applied manually (those lines are outside a diff hunk, so GitHub won't let me anchor a suggestion there):

```diff
     parser.add_argument(
-        "--no-streaming",
-        dest="streaming",
-        action="store_false",
-        help="Choose whether you'd like to download the entire dataset or stream it during the evaluation.",
+        "--streaming",
+        action="store_true",
+        help="Stream the dataset lazily over the network instead of downloading it in full before the evaluation. Off by default for reproducible benchmark timings.",
     )
```

Same two changes apply in `transformers/run_eval_ml.py` (lines 315-320 for the rename, line 357 for the `set_defaults` removal).


The inline suggestion below just removes the dead `set_defaults` line:

```suggestion
    args = parser.parse_args()
```

normalizer/data_utils.py
ebezzam: because I used a different name on separate sets 🙊

transformers/run_eval.py
```diff
@@ -67,22 +167,26 @@ def benchmark(batch, min_new_tokens=None):
...
+        inputs = inputs.to(args.device, dtype=torch_dtype)
 
         # 2. Model Inference
-        with sdpa_kernel(SDPBackend.MATH if args.torch_compile else SDPBackend.FLASH_ATTENTION):
+        if args.torch_compile is not None:
+            sdpa_backends = [SDPBackend.MATH]
+        else:
+            sdpa_backends = [SDPBackend.FLASH_ATTENTION, SDPBackend.EFFICIENT_ATTENTION, SDPBackend.MATH]
+        with sdpa_kernel(sdpa_backends):
+        # with sdpa_kernel(SDPBackend.MATH if args.torch_compile else SDPBackend.FLASH_ATTENTION):
```
Deep-unlearning: Leftover commented-out version of the `sdpa_kernel(...)` call that's already been replaced by the active `with sdpa_kernel(sdpa_backends):` on the line above. Safe to delete.

```suggestion
        with sdpa_kernel(sdpa_backends):
```

transformers/README.md
```diff
@@ -0,0 +1,111 @@
...
+To select a specific GPU (e.g. GPU 1):
+
+```bash
+docker run --gpus '"device=1"' \
+    -v $(pwd):/app \
+    -v $HF_HOME:/root/.cache/huggingface \
+    open-asr-transformers run_whisper.sh
+```
+
+### Run interactivately
```
Deep-unlearning: Typo: `interactivately` → `interactively`. (Also on the next line: `enter the contain` → `enter the container`.)

```suggestion
### Run interactively
```

Deep-unlearning (commented):
Follow-up thought (not blocking): `phi/` could also be folded into this standardized `run_eval.py` in a second pass. Phi uses beam search and a custom `MultipleTokenBatchStoppingCriteria`, neither of which the unified dispatch currently handles — so migrating it would force us to add `--num_beams` as a first-class arg and a clean way to register custom stopping criteria per model. That feels like a useful generalization (beam search isn't phi-specific; it's likely to come up again), so phi is a nice forcing function rather than an edge case. Happy to open a follow-up issue or PR.

Deep-unlearning:
We can add also cohere, should be low hanging!

Deep-unlearning:
@claude can you review this pr ?

transformers/run_eval.py
```diff
@@ -99,10 +204,34 @@ def benchmark(batch, min_new_tokens=None):
             pred_ids = pred_ids[:-padding_size, ...]
 
         # 3.2 Convert token ids to text transcription
-        pred_text = processor.batch_decode(pred_ids, skip_special_tokens=True)
```
ebezzam, line 102: `skip_special_tokens=True` would collapse double letters for CTC models!

abr/run_abr.sh
```diff
@@ -17,7 +17,7 @@ do
 
     python run_eval.py \
         --model_id=${MODEL_ID} \
-        --dataset_path="hf-audio/esb-datasets-test-only-sorted" \
+        --dataset_path="hf-audio/open-asr-leaderboard" \
```
ebezzam: Also took the opportunity to update the name of the dataset

ctranslate2/run_eval.py
```diff
@@ -136,6 +135,5 @@ def benchmark(batch):
         help="Number of warm-up steps to run before launching the timed runs.",
     )
     args = parser.parse_args()
-    parser.set_defaults(streaming=False)
```
ebezzam, line 139: Fixing streaming logic for other models as well

abr/run_eval.py
```diff
@@ -16,6 +16,7 @@ def main(args):
         args.model_id, trust_remote_code=True
     ).cuda()
     model = AutoModel.from_pretrained(args.model_id, trust_remote_code=True).cuda()
+    print(f"Model size: {sum(p.numel() for p in model.parameters()) / 1e9:.2f}B parameters")
```
ebezzam, line 26: Added model size print for all models. So that we can double check against what they report in the model name (which is sometimes only for the LLM)

normalizer/english_abbreviations.py
```diff
@@ -1741,3 +1741,168 @@
   "yoghurts": "yogurts"
 }
 
+
+english_name_normalizer = {
```
ebezzam, line 1748: Added mappings for names that sound similar but that could be mapped differently.

normalizer/normalizer.py
```diff
@@ -525,6 +525,58 @@ def __call__(self, s: str):
...
+    """
+    Collapse sequences of single-character tokens (letters or digits) into single words.
+
+    This normalizes acronym spacing so that both spaced-out and joined forms match:
+        - "b b c" -> "bbc"
+        - "5 g" -> "5g"
+
+    Lone single-character words surrounded by multi-character words are left untouched
+    (e.g. "a big cat" stays "a big cat").
+    """
```
ebezzam, line 538: added normalizer for acronyms as model/datasets may transcribe them differently

normalizer/normalizer.py
```diff
@@ -585,6 +637,20 @@ def __init__(self, english_spelling_mapping=english_spelling_normalizer):
...
+            r"\bwi\s+fi\b": "wifi",
+            r"\bhi\s+fi\b": "hifi",
+            r"\blo\s+fi\b": "lofi",
+            r"\bsci\s+fi\b": "scifi",
+            r"\be\s+mail\b": "email",
+            r"\be\s+book\b": "ebook",
+            r"\be\s+commerce\b": "ecommerce",
+            r"\bx\s+ray\b": "xray",
+            r"\bt\s+shirt\b": "tshirt",
+        }
```
ebezzam: to address common compounds, in case model/datasets transcribe them differently

transformers/run_eval.py
```diff
@@ -2,61 +2,163 @@
...
+    has_transcription_processor = hasattr(processor, "apply_transcription_request")
+
+    # Optional prompt for audio language models, newer models should use `apply_transcription_request`
+    text = None
+    if "granite-speech-3.3" in args.model_id.lower():
+        # create text prompt
+        chat = [
+            {
+                "role": "system",
+                "content": "Knowledge Cutoff Date: April 2024.\nToday's Date: December 19, 2024.\nYou are Granite, developed by IBM. You are a helpful AI assistant",
```
ebezzam, line 99: Leaving prompt that was contributed by model developer

transformers/run_eval.py
```diff
@@ -99,10 +204,34 @@ def benchmark(batch, min_new_tokens=None):
...
+                        pred_text.append(decoded[0] if isinstance(decoded, list) else decoded)
+                    except Exception as sample_error:
+                        print(f"Sample {i} decoding failed with error: {sample_error}. Setting to empty transcript.")
+                        pred_text.append("")
+        elif has_transcription_processor or texts is not None:
+            # Strip input prompt tokens
+            pred_text = processor.batch_decode(pred_ids[:, prompt_len:], skip_special_tokens=True)
+        elif is_ctc:
+            # don't use skip_special_tokens as it collapses double letters
+            pred_text = processor.batch_decode(pred_ids)
```
ebezzam, line 321: Important that CTC models DO NOT set `skip_special_tokens=True`

ebezzam (commented):
Some more comments on latest changes.

I think it's best to do Cohere and Phi4 separately as the current scripts are using remote code (namely not the "official" Transformers checkpoint and code). And it would be best to do them separately to ensure there isn't regression.

ebezzam:
@claude can you review this PR?

normalizer/normalizer.py
Deep-unlearning: I think it should be fine but they are some edge cases whenever there are spelling letter in the audio.

Deep-unlearning:
> Some more comments on latest changes.
> 
> I think it's best to do Cohere and Phi4 separately as the current scripts are using remote code (namely not the "official" Transformers checkpoint and code). And it would be best to do them separately to ensure there isn't regression.

Cohere does not use remote code, the current benchmarks is based on https://github.com/julianmack-cohere/open_asr_leaderboard/pull/1, with the only modification being that we drop the revision="refs/pr/11" argument when loading the processor and model 
However Cohere's transcribe flow diverges from the standard pattern in a few places, so folding it into the unified run_eval.py would need dedicated branches like we have for voxtral:

- Processor call takes extra kwargs: language=... and punctuation=(language != "en").
- Decode uses processor.decode (not batch_decode) with audio_chunk_index from the processor output + language.                                                                                                                                                                                     

(cf https://github.com/julianmack-cohere/open_asr_leaderboard/pull/1/changes#diff-0847fa25d67f8ce1c7df16b6f4f5737cad671582ffab42e4a906dc66f7578ddcR30)

Let's do this as a follow-up PR, I've already got something I can open once this is merged!

transformers/run_eval.py
Deep-unlearning: Side thought, not blocking: as we add more models with custom prompts / processor kwargs / generate kwargs (granite, voxtral, soon cohere), might be cleaner to move those into the .sh scripts rather than if "model_name" in args.model_id branches in run_eval.py.

transformers/run_eval.py
Deep-unlearning: Thanks for that important fix!

Deep-unlearning (approved):
LGTM! Left a few non-blocking nits and some follow-up thoughts for Cohere

ebezzam:
@Deep-unlearning Thanks for pointing out the new Cohere changes! I've incorporated its English short form into the Transformers folder. 

I also did something tweaks to normalization and post-processing:
- expand filler words to remove
- better handling of acronyms

Below are results on a few models to show how the new normalization utilities affect results on a few models. 

Overall it actually helps them! Gigaspeech WER is notably better for all of them. For some combos, the WER gets slightly worse, the gains still outweight that.

| Model  | Avg WER | AMI | E22 | GS | LS-C | LS-O | SPG | TED | VP  | 
|-------|---------|-----|-----|-----|------|------|-----|-----|-----|
| Cohere (current) | **5.42%** | 8.12| 10.86 | 9.34 | 1.25 | 2.37 | 3.08 | 2.49 | 5.87 | 
|  Cohere ([above](https://github.com/julianmack-cohere/open_asr_leaderboard/pull/1/changes#diff-0847fa25d67f8ce1c7df16b6f4f5737cad671582ffab42e4a906dc66f7578ddcR30) + updated normalizer) | **5.32%** |   8.16 | 10.78 | 8.64 | 1.24 | 2.34 | 3.04 | 2.48 | 5.86 | 
|  Cohere (standardized Transformers, stays same!) |  **5.32%** |   8.16 | 10.78 | 8.64 | 1.24 | 2.34 | 3.04 | 2.48 | 5.86 | 


| Model  | Avg WER | AMI | E22 | GS | LS-C | LS-O | SPG | TED | VP  | 
|-------|---------|-----|-----|-----|-----|------|------|-----|-----|
| Whisper Turbo v3 (current) | **7.82%** |  16.13 | 11.63 | 10.14 | 2.01 | 4.24 | 2.97 | 3.57 | 11.87 |
|  standardized + updated normalizer | **7.46%**  | 15.93 | 11.09 | 9.62 | 1.99 | 4.27 | 3.02 | 3.8 | 9.96 |


And below models which aren't affected by the standardization but by the updated normalizer

| Model  | Avg WER | AMI | E22 | GS | LS-C | LS-O | SPG | TED | VP  | 
|-------|---------|-----|-----|-----|-----|------|------|-----|-----|
| Qwen 1.7B (current) | **5.75625%**  |  10.56 | 10.25 | 8.74 | 1.63 | 3.4 | 2.84 | 2.28 | 6.35 || 
| with updated normalized | **5.54%** |  9.81 | 10.21 | 7.88 | 1.61 | 3.4 | 2.82 | 2.29 | 6.3 | 

| Model  | Avg WER | AMI | E22 | GS | LS-C | LS-O | SPG | TED | VP  | 
|-------|---------|-----|-----|-----|-----|------|------|-----|-----|
| Qwen 0.6B (current) | **6.42375%** | 11.66 | 11.06 | 9.14 | 2.13 | 4.45 | 3.03 | 2.85 | 7.07 | 
| with updated normalized | **6.225%** | 11.06 | 11.02 | 8.27 | 2.12 | 4.45 | 3.01 | 2.81 | 7.06| 


I didn't add Cohere multilingual and longform because:
- multilingual were getting slightly worse results
- longform requires this custom chunking for the audio to fit in memory, which I think wouldn't be in the spirit of a standard Transformers approach

Let's do those in a separate PR.

normalizer/english_abbreviations.py
```diff
@@ -1,4 +1,5 @@
 english_spelling_normalizer = {
+  "ok": "okay",
```
nithinraok, line 2: can we also add 'kay -> okay

nithinraok: and can we also include abbrevations normalization. 

For example: 
`L.C.D` -> `LCD`
`L. C. D.` -> `LCD` 
`L C D` -> `LCD` etc

normalizer/eval_utils.py
```diff
@@ -99,6 +100,30 @@ def write_manifest(
             f"The number of samples in `transcription_time` ({len(transcription_time)}) "
             f"must match `references` ({len(references)})."
         )
+    if audio_filepaths is not None and len(audio_filepaths) != len(references):
+        raise ValueError(
+            f"The number of samples in `audio_filepaths` ({len(audio_filepaths)}) "
+            f"must match `references` ({len(references)})."
+        )
+
+    # Filter out samples where the normalized reference is empty,
```
nithinraok: why to skip? This can be good case for hallucination testing, when reference is empty?

normalizer/normalizer.py
```diff
@@ -585,6 +639,23 @@ def __init__(self, english_spelling_mapping=english_spelling_normalizer):
         }
         self.standardize_numbers = EnglishNumberNormalizer()
         self.standardize_spellings = EnglishSpellingNormalizer(english_spelling_mapping)
+        self.standardize_names = EnglishNameNormalizer()
+        self.standardize_acronyms = EnglishAcronymNormalizer()
+        # Hardcoded compound words that become two tokens after hyphen/symbol removal
+        self.compound_words = {
```
nithinraok: Few more:
`al right` -> `alright` 
`all right` -> `alright` 
`how ever` -> `however`

normalizer/english_abbreviations.py
nithinraok: `et cetra` -> `etc`

normalizer/normalizer.py
nithinraok: Can we leverage this option from editdistance? https://github.com/pzelasko/kaldialign/pull/29 so all compound words are captured without hardcoding

normalizer/english_abbreviations.py
ebezzam: abbreviation normalization is now handled with [this](https://github.com/ebezzam/open_asr_leaderboard/blob/0d1055ce2ed797c825e7881f8ae8b807c928a567/normalizer/normalizer.py#L528), so that all those examples would map to lowercase `lcd`

and added the others!

normalizer/eval_utils.py
ebezzam: good point. this was causing issues when I had reference audio that just had filler words, and I think the WER computation was trying to normalize by number of words in the reference (=0) which led to errors. But I guess those samples could rather be given a WER of 100%? or do we let WER go above 100%?

EDIT: removed that skip and added [this logic ](https://github.com/ebezzam/open_asr_leaderboard/blob/6962cd21d9ef3cb4513ea1d0b32ba58b3abd7e08/normalizer/eval_utils.py#L225-L227)to penalize hallucinated insertions. wdyt?

ebezzam: also I had an idea to avoid computing model outputs whenever we change the normalizer: the JSONL files should store **unnormalized** reference and model outputs, and normalization is shifted to within `score_results`.

normalizer/normalizer.py
ebezzam: added [here](https://github.com/ebezzam/open_asr_leaderboard/blob/6962cd21d9ef3cb4513ea1d0b32ba58b3abd7e08/normalizer/eval_utils.py#L219-L235), just for English for now.

normalizer/eval_utils.py
```diff
@@ -174,30 +188,52 @@ def parse_filepath(fp: str):
...
+                    # empty reference: every predicted word is an insertion
+                    total_ins += len(pred_words)
+                else:
+                    r = kaldi_edit_distance(ref_words, pred_words, merge_compounds=True)
+                    total_ins += r["ins"]
+                    total_del += r["del"]
+                    total_sub += r["sub"]
+                    total_ref_words += r["ref_len"]
+            total_errors = total_ins + total_del + total_sub
+            wer = total_errors / total_ref_words if total_ref_words > 0 else 0.0
```
ebezzam: @pzelasko I've incorporated `kaldialign` here! Could I get your feedback on usage and if there's a recommended WER method to use with the output of `kaldi_edit_distance`?

PR #167, opened 2026-06-17
Correct abbreviation for 'flyer / flier'

kintel:
The spelling mapping for flyer -> flier was copies as the verbatim "flyer / flier" and "flier / flyer" strings. The spaces and slashes will cause this to never be matched by the normalizer.

Changing to "flyer: flier" to allow the matcher to work as intended.


ebezzam:
thanks for catching that!

ebezzam (approved):

PR #185, opened 2026-07-14
Add reusable multilingual API benchmark workflow

MyButtermilk:
## Summary

- add an explicit-dispatch and reusable GitHub Actions workflow for all registered multilingual API models
- centralize the 13 benchmark configurations, leaderboard column mapping, artifact selection, and scoring
- add deterministic checkpoint/resume and prior-run reuse so completed paid samples are not transcribed twice
- fail closed on missing samples/manifests instead of publishing partial scores
- document secret/concurrency setup and add secret-free CI coverage with a fake provider

## Safety and behavior

- defaults to a one-sample smoke test; `max_samples=0` is required for a full paid run
- never runs paid transcription from pull requests and does not use `pull_request_target`
- keeps workflow permissions read-only and emits reviewable artifacts instead of pushing leaderboard rows
- supports an explicitly passed reusable-workflow secret or a maintainer-selected repository/environment secret
- preserves API RTFx convention (`-1`) while retaining per-sample duration/timing data in manifests

## Validation

- `python -m unittest discover -s tests -v` (11 tests)
- `ruff check api/multilingual_benchmark.py api/run_eval_ml.py tests/test_multilingual_benchmark.py tests/test_run_eval_ml_resume.py`
- `actionlint` v1.7.12 on both added workflows

The paid workflow itself is intentionally manual and can only be dispatched after the workflow exists on a repository's default branch. Pull-request CI exercises the generic orchestration, scorer, ordering, normalization, checkpoint, and reuse logic without credentials or provider calls.


ebezzam:
@MyButtermilk thanks for this new workflow! definitely going to take a look, although there are first some updates to the multilingual workflow (like in the English track) that we'll address before looking into this.

MyButtermilk:
@ebezzam : Any update?

ebezzam:
@MyButtermilk thanks for the ping! I'll try to take a look in two weeks when back from holiday! I need more time to look into but I'm interested as we can improve the workflow for API models. 

Can you sync with main in the meantime?

MyButtermilk:
@ebezzam  done.

PR #202, opened 2026-08-18
Add support for AA-Earnings-Cleaned set

nithinraok:
## Description

  Adds [`ArtificialAnalysis/Earnings22-Cleaned-AA-chunked`](https://huggingface.co/datasets/ArtificialAnalysis/Earnings22-Cleaned-AA-chunked) as an evaluation set, plus the generic support it needs.

  This set is structurally different from every other set on the leaderboard: it ships **one reference transcript per session, not per audio row**. It is 341 VAD-derived chunks (≤30 s) cut from 6 earnings calls, and the chunked repo carries no transcripts at all — the references live in the parent repo [`ArtificialAnalysis/Earnings22-Cleaned-AA`](https://huggingface.co/datasets/ArtificialAnalysis/Earnings22-Cleaned-AA), one per call. So WER has to be computed after reassembling each
  call from its chunks.

  Rather than special-casing this dataset, the PR adds a small generic path for "chunked" datasets:

  - `normalizer/data_utils.py`: a `CHUNKED_DATASETS` registry mapping a chunked dataset repo to the parent repo holding its transcripts. `load_data()` dispatches to `load_chunked_data()`, which fetches the audio (it isn't referenced by the metadata file) and attaches each chunk's parent transcript as `text`.
  - `normalizer/eval_utils.py`: `write_manifest()` gains an optional `extra_fields`, used to carry `parent_id` / `chunk_index` into the results manifest. `merge_chunked_manifest()` groups rows by `parent_id`, concatenates predictions in `chunk_index` order, and sums durations and times; `score_results()` applies it before scoring. Manifests without `parent_id` pass through untouched, so nothing changes for existing datasets.
  - Because the chunk metadata travels in the manifest, results stay order-independent — harnesses that re-sort by duration for batching (e.g. NeMo) work unmodified.
  - Wired through `nemo_asr/run_eval.py` and `transformers/run_eval.py`, and added as a column in the public CSV summary.

  ### How to run

  ```bash
  cd nemo_asr
  python run_eval.py \
      --model_id=nvidia/parakeet-tdt-0.6b-v3 \
      --dataset_path="ArtificialAnalysis/Earnings22-Cleaned-AA-chunked" \
      --dataset="earnings22_cleaned_aa_chunked" \
      --split="test" \
      --device=0 --batch_size=32 --max_eval_samples=-1
  ```

  `nemo_asr/run_parakeet.sh` now takes an optional third field per dataset config (`"name split [dataset_path]"`, defaulting to `hf-audio/open-asr-leaderboard`) and includes this set, so `bash run_parakeet.sh` covers it.

  ### Results

  | model | Earnings22 (current, original refs) | Earnings22-Cleaned-AA-chunked | RTFx |
  | -- | -- | -- | -- |
  | `nvidia/parakeet-tdt-0.6b-v2` | 10.78 | 6.79 | 1824 |
  | `nvidia/parakeet-tdt-0.6b-v3` | 10.77 | 5.86 | 1715 |

  Earnings22 column is the published number from `english_short_latest.csv`. RTFx is from local runs on a single A6000 at batch 32 — **not** leaderboard-grade (H200, max batch size, `torch.compile`); please re-measure via HF Jobs before publishing any throughput figure.

  ### Why this set is worth adding

  The cleaning matters more than the usual "another test set" argument. Controlled check — same audio, same model, same predictions, only the reference swapped (`parakeet-tdt-0.6b-v2`, on the 5 of 6 calls where AA did not also trim the audio):

  - vs. original Earnings22 transcripts: **11.31 %** WER
  - vs. AA cleaned transcripts: **7.42 %** WER

  So ~3.9 p.p. of what the leaderboard currently attributes to model error on Earnings22 is error in the reference transcript. That is consistent with AA's own reported 5.6 p.p. average across models.

  ### Notes for reviewers

  - **Aggregation.** `score_results()` pools errors across the 6 session rows (i.e. word-count weighted), which is what this repo does for every other set. AA instead documents an audio-duration-weighted average across sessions. Kept the repo's convention for consistency — happy to switch if you'd rather match AA exactly.
  - **Merging.** Chunks are non-overlapping and cover each call in order, so predictions are joined in `chunk_index` order with a single space; no overlap/dedup logic is needed.
  - **Partial runs.** `merge_chunked_manifest()` warns when a session's chunk indices aren't contiguous from 0 (e.g. under `--max_eval_samples`), since partial sessions silently inflate deletions.
  - This set is 6 calls, versus the 125-call original Earnings22 test set, so the two WER columns above are not a like-for-like subset. The controlled experiment is the clean evidence for the cleaning effect specifically.

  ## Type of change
  - [ ] New model
  - [x] New dataset
  - [ ] Bug fix
  - [x] New feature
  - [ ] Other

  ## New Dataset Checklist
  - [x] The dataset is hosted on the HF Hub with **just** the test set. (341 rows, `test` only; parent transcript repo is likewise `test` only, 6 rows. Both are hosted by Artificial Analysis.)
  - [ ] Create a new Bash script with one of the existing models.

    Added the dataset to the existing `nemo_asr/run_parakeet.sh` instead of creating a new script, since it's an additional set for existing models rather than a new model library. Let me know if you'd prefer a standalone script.

### DIFF
--- a/abr/submit_jobs.sh
+++ b/abr/submit_jobs.sh
@@ -5,12 +5,33 @@
 # ── Configuration ────────────────────────────────────────────────────────────
 SPACE="${SPACE:-hf-audio/open-asr-leaderboard-abr}"
 RESULTS_BUCKET="${RESULTS_BUCKET:-hf-audio/asr_leaderboard_h200}"
-DATASET_PATH="${DATASET_PATH:-hf-audio/open-asr-leaderboard}"
+DEFAULT_DATASET_PATH="${DEFAULT_DATASET_PATH:-hf-audio/open-asr-leaderboard}"
 FLAVOR="${FLAVOR:-h200}"
 ORG_NAME="${ORG_NAME:-}"
 BATCH_SIZE=512
 WARMUP_STEPS=5
 SUBBATCH_SAMPLES=30000000
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+# Set USE_LOCAL_SCRIPT=1 to run your local run_eval.py instead of the version
+# committed to the Space (useful for iterating without pushing to the Space).
+USE_LOCAL_SCRIPT="${USE_LOCAL_SCRIPT:-1}"
+LOCAL_SCRIPT_INJECT=""
+if [[ "$USE_LOCAL_SCRIPT" == "1" ]]; then
+    RUN_EVAL_B64=$(base64 -w0 "${SCRIPT_DIR}/run_eval.py")
+    LOCAL_SCRIPT_INJECT="echo '${RUN_EVAL_B64}' | base64 -d > /app/run_eval.py &&"
+fi
+
+# Set USE_LOCAL_NORMALIZER=1 to inject your local normalizer/ package into the
+# job (so normalizer changes take effect without updating the HF Space).
+USE_LOCAL_NORMALIZER="${USE_LOCAL_NORMALIZER:-1}"
+LOCAL_NORMALIZER_INJECT=""
+if [[ "$USE_LOCAL_NORMALIZER" == "1" ]]; then
+    NORMALIZER_B64=$(tar --exclude='__pycache__' --exclude='*.pyc' -czf - -C "${REPO_ROOT}" normalizer | base64 -w0)
+    LOCAL_NORMALIZER_INJECT="echo '${NORMALIZER_B64}' | base64 -d | tar -xzf - -C /app &&"
+fi
 
 # ── Models: "model_id revision" ──────────────────────────────────────────────
 MODEL_CONFIGS=(
@@ -18,10 +39,11 @@ MODEL_CONFIGS=(
     "abr-ai/niagara-38m-batch.en 4f3ec18d377b1fd01e94d15dc9b9db0a8cd74bd2"
 )
 
-# ── Datasets: "name split" ────────────────────────────────────────────────────
+# ── Datasets: "name split [dataset_path]" ─────────────────────────────────────
+# dataset_path defaults to $DEFAULT_DATASET_PATH when omitted.
 DATASET_CONFIGS=(
     "ami_cleaned test"
-    "earnings22 test"
+    "earnings22_cleaned_aa_chunked test ArtificialAnalysis/Earnings22-Cleaned-AA-chunked"
     "gigaspeech_cleaned test"
     "librispeech test.clean"
     "librispeech test.other"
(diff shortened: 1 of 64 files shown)


abr/submit_jobs.sh
```diff
@@ -5,23 +5,45 @@
...
+fi
+
+# Set USE_LOCAL_NORMALIZER=1 to inject your local normalizer/ package into the
+# job (so normalizer changes take effect without updating the HF Space).
+USE_LOCAL_NORMALIZER="${USE_LOCAL_NORMALIZER:-1}"
+LOCAL_NORMALIZER_INJECT=""
+if [[ "$USE_LOCAL_NORMALIZER" == "1" ]]; then
+    NORMALIZER_B64=$(tar --exclude='__pycache__' --exclude='*.pyc' -czf - -C "${REPO_ROOT}" normalizer | base64 -w0)
+    LOCAL_NORMALIZER_INJECT="echo '${NORMALIZER_B64}' | base64 -d | tar -xzf - -C /app &&"
+fi
```
ebezzam, line 34: adding this logic in all of the submit job scripts -> to inject the local normalizer and run_eval script into the space (so that the space/docker container doesn't need to be rebuilt for updates)

normalizer/eval_utils.py
```diff
@@ -222,36 +313,45 @@ def parse_filepath(fp: str):
...
+                ),
             },
         ),
         (
             "public",
-            None,   # always printed when public datasets are present
+            None,  # always printed when public datasets are present
             "model,RTFx,License,Size (B),# Languages,Encoder,Decoder,"
-            "AMI-Cleaned WER,Earnings22 WER,Gigaspeech-Cleaned WER,LS Clean WER,LS Other WER,SPGISpeech WER,Voxpopuli-Cleaned-AA WER",
+            "AMI-Cleaned WER,Earnings22-Cleaned-AA-chunked WER,Gigaspeech-Cleaned WER,LS Clean WER,LS Other WER,SPGISpeech WER,Voxpopuli-Cleaned-AA WER",
```
ebezzam, line 332: replacing old earnings22 with this cleaner version

nemo_asr/run_parakeet.sh
```diff
@@ -35,16 +37,18 @@ DATASET_CONFIGS=(
     "librispeech test.clean"
     "librispeech test.other"
     "spgispeech test"
+    "earnings22_cleaned_aa_chunked test ArtificialAnalysis/Earnings22-Cleaned-AA-chunked"
```
ebezzam, line 40: it could be nice to have this chunked dataset in the aggregated dataset along with the full transcript, but I'm fine with this approach. maybe we figure out a different approach in the future 🙂

 this is also good way to give credit/traffic to others who've prepared the datasets

ebezzam (approved):
Thanks for the assist @nithinraok! I've updated all the submit job scripts to evaluate on the cleaner Earnings22 split. Updated results on HF jobs with H200 can be found [here](https://huggingface.co/datasets/hf-audio/open-asr-leaderboard-results/discussions/16).

Issue #170, opened 2026-06-24
Potential Data Leakage Risk for API Models on Private Evaluation Sets


linyubupa:
Hi OpenASR team,

I have a concern regarding the private evaluation datasets.

For open-weight models, the training data is fixed and publicly released. However, API-based ASR services are continuously updated and may potentially collect evaluation traffic. Over time, repeated submissions of the same private benchmark samples could lead to indirect exposure of the private dataset, allowing providers to optimize for it intentionally or unintentionally.

One observation that motivated this question is the large improvement of Microsoft Azure Speech between the 05-2026 and 06-2026 releases on the private benchmark. Its average private WER improved from 10.10 to 8.08, with conversational WER improving from 18.18 to 14.63 in a single monthly release. Such improvements may be entirely legitimate, but they also highlight the difficulty of ensuring that a fixed private benchmark remains truly unseen for continuously updated API services.

If private benchmark samples are repeatedly evaluated against the same APIs over long periods, there is a risk that benchmark traffic could eventually influence future model updates, intentionally or unintentionally. In that case, the private benchmark would gradually lose its value as an unseen test set, and rankings might no longer reflect true generalization performance.

Has the leaderboard considered this issue? Are there any safeguards (e.g., dataset rotation, refreshes of the private evaluation sets, or agreements that evaluation traffic is not used for training) to prevent private data leakage for API models?

I am not accusing any specific provider of misconduct. Rather, I believe this is an important benchmarking challenge for all continuously updated closed-source ASR services.

Thanks.

<img width="3148" height="1224" alt="Image" src="https://github.com/user-attachments/assets/4f58c399-f65a-4271-a625-b29dd42c7b8a" />

<img width="1547" height="473" alt="Image" src="https://github.com/user-attachments/assets/586390be-bb8f-404f-96dc-04dad5ea618f" />

ebezzam:
Hi @linyubupa thank you for the comment and the valid concern!

Firstly, on the apparant "improvement" of `microsoft/azure-speech-06-2026`:
- The same endpoint was used in both cases, as pinned [here](https://github.com/huggingface/open_asr_leaderboard/blob/4e4880b9fb203d60830ed2920c521e067026f269/api/providers/microsoft_azure_provider.py#L14). The date change is to reflect when we evaluated.
[...]

PR #204, opened 2026-08-22
Modulate English/Multilingual 

angeljcarrillo:
## Description

Updates the existing **`modulate/vfast`** entry on the **English short-form track**. Two changes:

1. **Improved transcription model.** Modulate's hosted English system has been improved since the merged entry. Re-running the English short-form suite shows gains on all seven datasets, with a composite WER of **4.19**. We re-ran the suite multiple times to confirm the gains reproduce rather than reflecting run-to-run variance.
2. **Endpoint migration.** The English vfast system now runs under the multilingual endpoint, `velma-2-stt-batch-multilingual-vfast`. Both endpoints currently serve it and return identical results; `velma-2-stt-batch-english-vfast` will be retired at a future date, so the provider now points at the endpoint that will remain available.

This is **not a new model entry** — it updates the existing `modulate/vfast` row and should not add a second Modulate submission.

It is evaluated through Modulate's hosted endpoint (proprietary / API lane): `run_eval.py` loads each dataset, calls the provider for every clip, and computes WER; the provider uploads the clip to the endpoint and the endpoint returns a transcript. This PR changes only the API provider; it does not change `run_eval.py`, the Dockerfile, or the normalizer.

**How to run / reproduce (self-service — no key request needed):**

1. Create an account at **https://platform.modulate.ai/signin**
2. Generate an API key in the dashboard. New accounts receive free credit sufficient to cover a full run across all 7 datasets, with headroom to spare.
3. Export your key (the endpoint is set automatically by the provider — no endpoint configuration is required):

```bash
export MODULATE_API_KEY="<your generated key>"
```

4. Run `bash run_api.sh` (the `modulate/vfast` entry is already enabled in `MODEL_CONFIGS`).

Evaluated **out-of-the-box**: English, no prompt or per-dataset decoding overrides.

Changed files:

- `providers/modulate_provider.py` — point both variants at `velma-2-stt-batch-multilingual-vfast`, and declare the `language` form field for both. The endpoint requires the language to be declared to take the direct-routing path; omitting it falls back to per-segment language ID. Defaults to `en`.

API reference: https://docs.modulate.ai/api-reference/stt/batch-multilingual-vfast

Could you also update the API reference link on the leaderboard entry to the URL above? The current link points at the endpoint being retired.

## Type of change

- [ ] New model
- [ ] New dataset
- [ ] Bug fix
- [ ] New feature
- [x] Other (updated results + endpoint migration for an existing entry)

## Checklist

- [x] Report results (WER per split, average WER, RTFx) — see below.
  - The HF-Hub `.eval_results/open_asr_leaderboard.yaml` step is **N/A**: `modulate/vfast` is a proprietary API with no model repo on the HF Hub.

### HF Jobs evaluation (recommended)

This is an **API-lane** entry: the model is a hosted endpoint, so there is no local checkpoint, Dockerfile, or `run_eval.py` change to package into a Space. Reproduction is fully self-service against the public endpoint (see the run steps in the Description). Happy to prepare/duplicate a Space or provide a key directly if the maintainers prefer that route for HF Jobs verification.

- [ ] (Custom Space) — N/A for this API entry (no custom dependencies/Dockerfile).
  - [ ] Modify `Dockerfile` — N/A (no local model).
  - [ ] Adapt `run_eval.py` — N/A (uses the existing API `run_eval.py`, unchanged).
  - [ ] `trust_remote_code` / `revision` — N/A (API model, no remote code).
- [ ] Model-library folder + `submit_jobs.sh` — N/A (uses the existing API run script; `modulate` is an existing provider under the existing API path).

### Key guidelines

- [x] **Same decoding hyper-parameters across all datasets** — out-of-the-box, English, no prompt / keyterms / per-dataset overrides.
- [x] `run_eval.py` supports batch processing and uses the repo's data/normalizer utilities (unchanged).
- [ ] Maximum batch size on an H200 — N/A (hosted API; not run on a local GPU).
- [ ] `torch.compile` / warmup — N/A (hosted API).
- [x] **API model → API key:** verification is self-service — maintainers create an account at the link above and generate their own key; free credit covers the full public run and the private datasets, so no waiting on us. We can also provide a key directly if preferred.
- [ ] HF Space for HF Jobs reproduction — see the note above; happy to prepare one on request.
- [x] Results on the public sets + RTFx reported below.
- [x] Model metadata (unchanged):

| License | Size (B) | # Languages | Encoder | Decoder |
| --- | --- | --- | --- | --- |
| Proprietary | N/A (hosted API) | 1 (English) | N/A | N/A |

## Results — English short-form (cleaned splits)

Per-split WER and RTFx as reported by the harness (`score_results`) on this run, using the `ami_cleaned` / `gigaspeech_cleaned` / `voxpopuli_cleaned_aa` splits.

| Dataset | WER | RTFx |
| --- | --- | --- |
| LibriSpeech test.clean | 0.86 | 1.84 |
| LibriSpeech test.other | 1.75 | 1.78 |
| SPGISpeech | 2.06 | 2.02 |
| VoxPopuli (cleaned_aa) | 3.56 | 2.35 |
| AMI (cleaned) | 6.33 | 1.37 |
| GigaSpeech (cleaned) | 6.90 | 1.68 |
| Earnings22 | 7.88 | 1.52 |
| **Average (7 sets)** | **4.19** | **1.86** |

Run-to-run composite variance on this system is small (a few hundredths of a point); the reported average is **4.19**. Improvements reproduce across all seven datasets. Verified clean: all clips transcribed, no empty predictions.

> RTFx here is measured end-to-end through the hosted API under our serving concurrency for an async (pre-recorded) API, so — like other API entries — it is not a GPU-throughput figure and is not directly comparable to local open-weights RTFx on reference hardware. WER is the directly comparable metric.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -55,7 +55,7 @@ License | Size (B) | # Languages | Encoder | Decoder | (Recommended) Link to tra
 - [ ] Provide a link to your model's documentation that we can link on the leaderboard.
 - [ ] Please provide the following information. Note that the cost should represent the "entry-level" costs, e.g. when one first signs up to your platform.
 
- # Languages | Cost ($/hour) | Link to model announcement or API
+Cost ($/hour) |  # Languages | Link to model announcement or API
  -- | -- | -- 
  x | x | x 
 
--- a/README.md
+++ b/README.md
@@ -101,6 +101,11 @@ You can also specify your own model and its performance as such:
 
 ![Custom model](scripts/data/MY_MODEL_en_shortform_rtfx_wer.png)
 
+# Benchmark fitting
+
+The `benchmark_fitting` folder contains scripts to measure whether a model overfits reference transcripts in the test sets. See
+[`benchmark_fitting/README.md`](benchmark_fitting/README.md) for more information.
+
 # Contributing a model or dataset
 
 Please follow the [pull request template](./.github/PULL_REQUEST_TEMPLATE.md); it contains a submission checklist and guidelines.
--- a/README.md
+++ b/README.md
@@ -9,6 +9,9 @@ The Open ASR Leaderboard evaluates models on a diverse set of publicly available
 * **Main Test Sets (English, short-form):**
   The main benchmark datasets used for evaluation (short-form English) are available [here](https://huggingface.co/datasets/hf-audio/open-asr-leaderboard).
 
+* **Chunked sets with session-level references:**
+  [**Earnings22-Cleaned-AA-chunked**](https://huggingface.co/datasets/ArtificialAnalysis/Earnings22-Cleaned-AA-chunked) splits 6 earnings calls into 341 chunks, with one reference transcript per call in the parent [Earnings22-Cleaned-AA](https://huggingface.co/datasets/ArtificialAnalysis/Earnings22-Cleaned-AA) set. Models transcribe each chunk as usual; the scorer concatenates a call's chunk predictions before computing WER against the call-level reference. Run with `--dataset_path ArtificialAnalysis/Earnings22-Cleaned-AA-chunked --dataset earnings22_cleaned_aa_chunked --split test`.
+
 * **English, long-form:**
   The [**ASR Longform benchmark**](https://huggingface.co/datasets/hf-audio/asr-leaderboard-longform) dataset includes earnings21 and earnings22. We also evaluate on [CORAAL](https://huggingface.co/datasets/bezzam/coraal), but it is stored as a separate dataset since it has multiple splits.
 
--- a/abr/run_eval.py
+++ b/abr/run_eval.py
@@ -141,6 +141,8 @@ def main(args):
         remove_columns=["audio"],
     )
 
+    is_chunked = data_utils.is_chunked_dataset(args.dataset_path)
+
     all_results = {
         "audio_length_s": [],
         "transcription_time_s": [],
@@ -148,6 +150,8 @@ def main(args):
         "references": [],
         "audio_filepath": [],
     }
+    if is_chunked:
+        all_results.update({key: [] for key in data_utils.CHUNK_METADATA_KEYS})
     result_iter = iter(dataset)
     for result in tqdm(result_iter, desc="Samples..."):
         for key in all_results:
@@ -164,11 +168,22 @@ def main(args):
         audio_length=all_results["audio_length_s"],
         transcription_time=all_results["transcription_time_s"],
         audio_filepaths=all_results["audio_filepath"],
+        extra_fields={key: all_results[key] for key in data_utils.CHUNK_METADATA_KEYS}
+        if is_chunked
+        else None,
     )
     print("Results saved at path:", os.path.abspath(manifest_path))
 
-    norm_refs = [data_utils.normalizer(r) for r in all_results["references"]]
-    norm_preds = [data_utils.normalizer(p) for p in all_results["predictions"]]
+    if is_chunked:
+        sessions = data_utils.merge_chunked_manifest(data_utils.read_manifest(manifest_path))
+        references = [session["text"] for session in sessions]
+        predictions = [session["pred_text"] for session in sessions]
+    else:
+        references = all_results["references"]
+        predictions = all_results["predictions"]
+
+    norm_refs = [data_utils.normalizer(r) for r in references]
+    norm_preds = [data_utils.normalizer(p) for p in predictions]
     wer = wer_metric.compute(
         references=norm_refs, predictions=norm_preds
     )
--- a/abr/submit_jobs.sh
+++ b/abr/submit_jobs.sh
@@ -5,12 +5,33 @@
 # ── Configuration ────────────────────────────────────────────────────────────
 SPACE="${SPACE:-hf-audio/open-asr-leaderboard-abr}"
 RESULTS_BUCKET="${RESULTS_BUCKET:-hf-audio/asr_leaderboard_h200}"
-DATASET_PATH="${DATASET_PATH:-hf-audio/open-asr-leaderboard}"
+DEFAULT_DATASET_PATH="${DEFAULT_DATASET_PATH:-hf-audio/open-asr-leaderboard}"
 FLAVOR="${FLAVOR:-h200}"
 ORG_NAME="${ORG_NAME:-}"
 BATCH_SIZE=512
 WARMUP_STEPS=5
 SUBBATCH_SAMPLES=30000000
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+# Set USE_LOCAL_SCRIPT=1 to run your local run_eval.py instead of the version
+# committed to the Space (useful for iterating without pushing to the Space).
+USE_LOCAL_SCRIPT="${USE_LOCAL_SCRIPT:-1}"
+LOCAL_SCRIPT_INJECT=""
+if [[ "$USE_LOCAL_SCRIPT" == "1" ]]; then
+    RUN_EVAL_B64=$(base64 -w0 "${SCRIPT_DIR}/run_eval.py")
+    LOCAL_SCRIPT_INJECT="echo '${RUN_EVAL_B64}' | base64 -d > /app/run_eval.py &&"
+fi
+
+# Set USE_LOCAL_NORMALIZER=1 to inject your local normalizer/ package into the
+# job (so normalizer changes take effect without updating the HF Space).
+USE_LOCAL_NORMALIZER="${USE_LOCAL_NORMALIZER:-1}"
+LOCAL_NORMALIZER_INJECT=""
+if [[ "$USE_LOCAL_NORMALIZER" == "1" ]]; then
+    NORMALIZER_B64=$(tar --exclude='__pycache__' --exclude='*.pyc' -czf - -C "${REPO_ROOT}" normalizer | base64 -w0)
+    LOCAL_NORMALIZER_INJECT="echo '${NORMALIZER_B64}' | base64 -d | tar -xzf - -C /app &&"
+fi
 
 # ── Models: "model_id revision" ──────────────────────────────────────────────
 MODEL_CONFIGS=(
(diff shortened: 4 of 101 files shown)


angeljcarrillo:
Hi @ebezzam 👋 Hope you're doing well!

Back again — this one is an **update to the existing `modulate/vfast` entry**, not a new submission. Our hosted English system has improved since the merged entry, and re-running the English short-form suite gives a composite WER of **4.19**, with gains on all seven cleaned splits. I ran it a few times first to be sure it reproduces and isn't just run-to-run variance.

The English system also now runs under `velma-2-stt-batch-multilingual-vfast`. Both endpoints return identical results today, but the English-only one will be retired, so the provider points at the one that stays. Only `providers/modulate_provider.py` changed — and just to be explicit so it doesn't create extra work for you: this should update the existing `modulate/vfast` row, not add a second Modulate entry.

Reproduction is self-service as before (account at https://platform.modulate.ai/signin, free credit covers a full run) — or I'm very happy to send you a key directly, whatever is easier.

Also, thank you for the Docker setup you added for the API lane — it's a really nice improvement for keeping runs consistent. One small note from my side: it does need noticeably more memory, so running it directly on a laptop is a bit harder than the previous flow, which ran comfortably on just about any machine. Not a blocker at all, just sharing in case it's useful feedback.

One small ask whenever convenient: could the API reference link point to https://docs.modulate.ai/api-reference/stt/batch-multilingual-vfast? The current one points at the retiring endpoint.

Thanks again for your time — happy to re-run or adjust anything. 😃

angeljcarrillo:
Hi @ebezzam , hope you're doing well! Just checking in to see if anything is needed from my side on this one. No rush at all — looking forward to hearing from you whenever it's convenient. Thanks!

ebezzam (commented):
thanks @angeljcarrillo for the update info! 

How about leaving the `modulate_provider.py` file as is and rather setting `modulate/multilingual` in `run_api.sh` [here](https://github.com/huggingface/open_asr_leaderboard/blob/14469179567f4b829f96ba38f2f4651c52acc5ce/api/run_api.sh#L29) (and [here](https://github.com/huggingface/open_asr_leaderboard/blob/14469179567f4b829f96ba38f2f4651c52acc5ce/api/run_api.sh#L51) since it's mentioned)

such that the eval is launched like so:
```
MODEL="modulate/multilingual  25" \
bash api/run_api.sh
```

And then on the leaderboard, I'm thinking we also "retire" the `modulate/vfast` name, and have `modulate/multlingual` for both the English and multilingual tabs? This would be consistent with how other API models are presented (same name under English and multilingual)

angeljcarrillo:
That sounds good to me @ebezzam , and I agree with the reasoning. Leaving `modulate_provider.py` as is and setting `modulate/multilingual` in `run_api.sh` is the cleaner option, and retiring `modulate/vfast` so the same name appears under both the English and multilingual tabs is consistent with how the other API models are presented.

So it looks like we're aligned there. I really appreciate you looping back on this.

Separately, I noticed the dataset update and wanted to say I think it's a great addition. Having new datasets come in is genuinely valuable, since it surfaces where models are still weak rather than letting them settle on a fixed benchmark.

On that note, could I ask to wait until tomorrow? I'd like to run our numbers on the new dataset first, and then I can come back with the updated results so you can re-run them on your end to verify.

That did raise a process question I was curious about. When a new dataset is introduced, it looks like the existing entries aren't re-evaluated across the whole suite — only the new dataset gets run. I imagine that's because re-running everything would mean extra cost on your side, which completely makes sense. If that's right, would it be reasonable for submitters to proactively check and request updated numbers for their own models when this happens? I just wanted to check whether that matches how you're thinking about it.

ebezzam:
@angeljcarrillo great that you agree with keeping only `modulate/multilingual`

> On that note, could I ask to wait until tomorrow? I'd like to run our numbers on the new dataset first, and then I can come back with the updated results so you can re-run them on your end to verify.

yes I can wait for your green light!

> That did raise a process question I was curious about. When a new dataset is introduced, it looks like the existing entries aren't re-evaluated across the whole suite — only the new dataset gets run. I imagine that's because re-running everything would mean extra cost on your side, which completely makes sense. If that's right, would it be reasonable for submitters to proactively check and request updated numbers for their own models when this happens? I just wanted to check whether that matches how you're thinking about it.

One reason is cost, but the main one is to avoid any potential re-training that may occur from audio that is passed to the APIs. As we are not sure if they use the user audio to improve the performance of the models (related discussion https://github.com/huggingface/open_asr_leaderboard/issues/170). So we'd rather report the "fresh" first-time performance. 

But as you mentioned if there's a significant update / new version (as has happened for previous API models), we are willing to run the whole suite. Hope that makes sense!

angeljcarrillo:
Hi @ebezzam, thanks for waiting. Green light from our side. 🚀

We've already applied your suggestion on our branch — `run_api.sh` now refers to `modulate/multilingual` and the provider is back to upstream's version. The results below came from that setup, so it's tested and working end to end.

We re-ran the benchmark against our multilingual endpoint with the updated dataset included. Composite result:

`modulate/multilingual` — **composite WER 3.84%**, RTFx 1.90

| Dataset | WER | RTFx |
|---|---|---|
| AMI-Cleaned | 6.33 | 1.63 |
| Earnings22-Cleaned-AA-chunked | 6.00 | 2.49 |
| GigaSpeech-Cleaned | 6.90 | 1.79 |
| LS Clean | 0.87 | 2.04 |
| LS Other | 1.77 | 1.98 |
| SPGISpeech | 2.06 | 1.98 |
| Voice-Arena-Monsoon | 3.25 | 1.52 |
| Voxpopuli-Cleaned-AA | 3.56 | 2.05 |

On #170 — happy to be probed as deeply as you'd like, and glad to help however is useful. Briefly, our methodology: we fine-tune on the train splits of the benchmark datasets plus our own training data, and what we add beyond that is collected for weaknesses we've identified ourselves, focusing on our model, mostly from what customers run into rather than from the benchmark. Those weigh more for us, since the API is a product first. We then validate on internal benchmark sets of our own to confirm improvements generalize. Test splits are never used to fit any component of our system.

I've held that position for a long time, and not only because training on test data is cheating. It also destroys the one instrument that tells you whether generalization is real — you can ship something worse for customers while the leaderboard says it improved.

That is why I find a new private dataset most interesting: it's the one case where the model has to hold up on something we couldn't have prepared for. In the same spirit, would you consider weighting the private set more heavily in the composite? (I personally have the same feeling when we run our internal benchmark for generalization; we just do not know where it will land ) It's the one part of the suite submitters can't obtain or prepare for, since the references are never published, so it carries more signal about real generalization — and it pulls down anyone who has optimised for the public splits without you needing to make a judgement call about anyone.

If it helps to have one API provider go first on any of this, we're glad to be that. Part of what I'd like to see come out of this is a clear, repeatable path for API providers to keep their numbers current as their models change — right now it's unclear when to ask, and having a convention would make the board more accurate for everyone. Happy to help however is useful. 🙌🏻

ebezzam:
@angeljcarrillo thanks for the confirmation! I've launched the evals.

Thanks for sharing your methodology, glad to hear you prioritize real-world / customer performance 👏 your approach makes complete sense.

> would you consider weighting the private set more heavily in the composite?

it is a valid question and one that's come up elsewhere. We usually tell people that they can toggle which datasets they want or simply see the "Private" tab, but it's true that most people come to the leaderboard and just look at the default setting. So it may be something we do at one point, but still need to determine what would be the best weighting.

> If it helps to have one API provider go first on any of this, we're glad to be that. Part of what I'd like to see come out of this is a clear, repeatable path for API providers to keep their numbers current as their models change

thanks for the proposition! I think the best approach would be to have something automated, e.g. that runs the evaluations every week and syncs the results with the "dataset" where we read the results for the leaderboard page (namely [here](https://huggingface.co/datasets/hf-audio/open-asr-leaderboard-results/tree/main)). It would require setup a new workflow.

There was a workflow proposed in https://github.com/huggingface/open_asr_leaderboard/pull/185 with GitHub actions (_which I still haven't had the time to look into_), or one could be set up with [HF jobs](https://huggingface.co/docs/huggingface_hub/en/guides/jobs) (using CPU hardware). We'd have a preference for the latter, and (as a start) it could be a script that we manually run each week to launch the evals and sync with the [results dataset](https://huggingface.co/datasets/hf-audio/open-asr-leaderboard-results/tree/main). When that's working, we can then setup a way to automate every 1-2 weeks.

Not sure if you have the bandwidth to prototype something like that for your API? I had started a [submit_job script](https://github.com/huggingface/open_asr_leaderboard/blob/main/api/submit_jobs.sh) for API models (without syncing with the results dataset), as I generally still use the local `run_api.sh` as you've been using.

angeljcarrillo:
Thanks for the thorough follow-up, really appreciate you taking the time on this @ebezzam 

On automation: we agree HF Jobs is the better fit here, and we're glad to be the API provider that goes first. A clear, repeatable path for API providers to keep their numbers current as models change is exactly what we'd like to see come out of this too.

We'll loop back with a prototype. `submit_jobs.sh` already covers most of it, so from our side the gap looks like adding our provider's key and model entry, plus the sync step to the results dataset. A couple of things we'll likely need your input on as we go — whether the `open-asr-leaderboard-apis` Space picks up the provider automatically (the script injects `run_eval.py` and `normalizer/` but not `providers/`), and where results should land in the results dataset. We'll ask as we hit them rather than guess.

On credentials, we can commit to this now: we'll provide a dedicated API key for the leaderboard so the job can run whatever it needs, on whatever cadence you settle on. We'd rather you not be blocked waiting on us, and we're happy to put that trust in your hands. We'll scope the key to this use and size its quota for a weekly full-suite run, and we'll handle rotation on our side so it doesn't become maintenance for you.

Thanks again!

ebezzam:
thanks @angeljcarrillo for looking into a prototype for this!

> We'll loop back with a prototype. submit_jobs.sh already covers most of it, so from our side the gap looks like adding our provider's key and model entry, plus the sync step to the results dataset. A couple of things we'll likely need your input on as we go — whether the open-asr-leaderboard-apis Space picks up the provider automatically (the script injects run_eval.py and normalizer/ but not providers/), and where results should land in the results dataset. We'll ask as we hit them rather than guess.

I haven't used that [space](https://huggingface.co/spaces/hf-audio/open-asr-leaderboard-apis) in a while, I think the best would be to duplicate so that you can make tweaks and iterate on it. 

Similarly other components that you need for your protoypes:
- A [bucket](https://github.com/huggingface/open_asr_leaderboard/blob/e7cb8e9b79ab908302c30bdc9f38152ad666a43c/api/submit_jobs.sh#L17) for storing the results. It can be private which is what we do for all API model results. The public one looks like [this](https://huggingface.co/buckets/hf-audio/asr_leaderboard_h200).
- A results dataset like [this](https://huggingface.co/datasets/hf-audio/open-asr-leaderboard-results/tree/main) for writing the metrics into. For your prototype, you can similarly make a copy of this. (Note that the leaderboard page by default pulls the latest version of the CSV file [here](https://huggingface.co/spaces/hf-audio/open_asr_leaderboard/blob/main/init.py#L19).)

I've sent a LInkedIn invite so we can exchange more easily!

angeljcarrillo:
Thank you again for the pointer @ebezzam . I've already accepted the invitation, so we can follow up there.

In any case, once this branch is merged, I'll open a new branch off it to start that work and get it deployed. I appreciate the trust in us to move this forward.

Coming back to this branch — let me know when the scores are done. As mentioned, my main interest is how well we generalize. On the internal benchmark I collected for data generalization, I do see improvement, and in some cases essentially flawless results. Which raises a question: will this be the private benchmark we run again?

I'll leave that call to you. Personally, I think we should run it . I'd rather keep it fair game, even if that means we come out worse than our earlier model. Let's see.

Thanks again for your patience.

angeljcarrillo:
Hi @ebezzam ,

Apologies for this — we noticed your account had run out of credits, so we've gone ahead and added 1,000 to it.
Its very likely broke the scoring pipeline on your side. Please feel free to re-run everything so you get an accurate benchmark. Sorry again for the disruption and the delay this has caused.

Do let me know if anything else looks off.

ebezzam:
@angeljcarrillo thanks for the extra credits! here are the results I've obtained if you can confirm? avg WER is the same, and as before RTFx can be ignored

```
modulate/multilingual | ArtificialAnalysis-Earnings22-Cleaned-AA-chunked__test: WER = 6.00 %, RTFx = 1.27
modulate/multilingual | VoiceArena-Monsoon_en_IN_test__test: WER = 3.26 %, RTFx = 2.19
modulate/multilingual | hf-audio-open-asr-leaderboard_ami_cleaned_test: WER = 6.31 %, RTFx = 0.44
modulate/multilingual | hf-audio-open-asr-leaderboard_gigaspeech_cleaned_test: WER = 6.91 %, RTFx = 0.66
modulate/multilingual | hf-audio-open-asr-leaderboard_librispeech_test.clean: WER = 0.87 %, RTFx = 0.76
modulate/multilingual | hf-audio-open-asr-leaderboard_librispeech_test.other: WER = 1.77 %, RTFx = 1.71
modulate/multilingual | hf-audio-open-asr-leaderboard_spgispeech_test: WER = 2.06 %, RTFx = 1.93
modulate/multilingual | hf-audio-open-asr-leaderboard_voxpopuli_cleaned_aa_test: WER = 3.55 %, RTFx = 2.30

avg WER (modulate/multilingual) = 3.84
```

angeljcarrillo:
Thanks for running this, @ebezzam  — I can confirm these match on our end. 
Happy to help with anything else you need from our side to get this landed. In the meantime, I'll keep moving on the API-lane prototype we discussed.

ebezzam:
thanks for confirming! and thanks a bunch for working on that prototype, looking forward 🚀 
results may land tomorrow or early next week

ebezzam:
how many languages are supported by `modulate/multilingual`?

angeljcarrillo:
The endpoint supports 99 languages, so please do let us know if you'd like extra credits to test more of them — happy to top you up whenever it's useful. Per-language quality is tuned independently on our side, so results can vary from language to language rather than moving together.

I also owe you an update to the multilingual benchmark. I'd like to do some further work on a few specific languages first, so I can't promise a date yet, but I'll come back to you once I have something worth sharing. 😅

ebezzam:
results are live!

Issue #187, opened 2026-07-15
Add Dutch to the multilingual track — datasets already include nl, offering seed results


mrapaul:
Hi — the multilingual dataset (`nithinraok/asr-leaderboard-datasets`) already includes Dutch (`fleurs_nl`, `mcv_nl`, `mls_nl`), and `run_eval_ml.py` runs them as-is, but the board's displayed languages don't yet include `nl`.

We'd like to contribute Dutch as a benchmarked language. To make that concrete we've already run three models through this repo's own harness (`transformers` and `nemo_asr` `run_eval_ml.py`) on the Dutch configs, scored with `eval_utils.score_results(multilingual=True, language="nl")`:

| Model | fleurs_nl | mcv_nl | mls_nl |
|---|---|---|---|
| NVIDIA Canary-1b-v2 | 5.27 | 5.78 | 9.64 |
| NVIDIA Parakeet-TDT-0.6b-v3 | 6.50 | 5.47 | 11.27 |
| our Dutch model (Vlotspraak) | 4.63 | 4.30 | 5.11 |

We verified all three scored a byte-identical reference set per config (hashed the sorted refs; `n` and total duration match), and as a sanity check our Parakeet FLEURS-nl number reproduces NVIDIA's published figure to within ~0.2pp on a second harness.

Two things worth flagging up front: (a) our RTFx is far lower than the NeMo models' — we're a Whisper-large-v3 derivative run through the `transformers` backend, so that column is not flattering to us; (b) our weights aren't currently public, so our own row would be self-reported unless you'd rather run it yourselves — happy to arrange access if so.

Would you be open to a PR that (a) enables `nl` in the multilingual display/plot and (b) adds the Canary/Parakeet results to seed the column so it isn't a single-entry track? We're glad to run further models on the Dutch configs regardless of what happens with our own entry. What's the preferred path — HF Jobs, or a results PR?

mrapaul:
Small update: we've published a model card with the Dutch eval results and artifacts, so everything referenced above is now in one public place: https://huggingface.co/codesparktech/vlotspraak-dutch-asr

Since opening this issue we also ran the Common Voice 17 NL test set through the Dutch community benchmark's official sclite tooling (opensource-spraakherkenning-nl): 3.5% WER on the full 11,266-clip set vs 4.3% for the previous best there, with a same-model control reproducing the 4.3% exactly — submitted as opensource-spraakherkenning-nl/ASR_NL_results#7.

[...]

PR #208, opened 2026-08-25
Add Dutch (nl) to the multilingual evals

mrapaul:
Steps 1 and 2 from #187, following @ebezzam's outline. No results here, just the wiring, so you can run whichever models you want on your side.

### Why it's cheap

All three multilingual sets already ship an `nl` config in `hf-audio/open-asr-leaderboard-multilingual-datasets` (`fleurs_nl`, `mcv_nl`, `mls_nl`), so Dutch needs no new data. It gets the same three-column shape as fr/it/es.

### What changed

**Scoring**
- `normalizer/eval_utils.py`: registered the `ml_nl` family with FLEURS/MCV/MLS.
- `scripts/score_bucket_results.py`: added `nl` to `ML_LANGUAGES` (so `--language nl` is accepted and the default sweep covers it) and added the three `nl_*` columns to the combined multilingual CSV.

**Job scripts**
Added `fleurs nl` / `mcv nl` / `mls nl` to every `*_ml.sh` whose model declares Dutch: Whisper, Parakeet-TDT-0.6b-v3, Canary-1b-v2, Voxtral (mini / small / realtime), Cohere transcribe, Qwen3-ASR (both the toolkit and the `-hf` path), Phi-4-multimodal, VibeVoice-ASR, Nemotron streaming, omniASR, and the API providers. The local `run_*_ml.sh` runners got the same languages so they don't drift from the submit scripts.

`granite/submit_jobs_nar_ml.sh` is deliberately untouched: `ibm-granite/granite-speech-4.1-2b-nar` lists en/fr/de/es/pt only. Language support was taken from each model card rather than assumed.

**Two things that would have gone wrong quietly**
- `qwen/run_eval_ml.py` and `phi/run_eval_ml.py` map language codes to language names and pass the name to the model. Without an `nl` entry, `LANGUAGE_NAMES.get("nl")` returns `None`, so Dutch would have run on auto-detect while every other language was forced. Added `"nl": "Dutch"` to both.
- `api/providers/microsoft_azure_provider.py` maps codes to BCP-47 locales; added `nl-NL`.

### Checks

`num2words` supports `nl`, so digits normalize the way they do for the other languages (`21` becomes `eenentwintig`).

Scored a Dutch manifest end to end with `score_results(multilingual=True, language="nl", families=["ml_nl"])`, and the compound merging in `normalize_compound_pairs` behaves: `binnenstad` vs `binnen stad` costs 0 errors. That one matters more for Dutch than for most of the current languages, given how freely it compounds.

```
CSV Summary (nl):
model,RTFx,FLEURS WER,MCV WER,MLS WER
```

Also `bash -n` on all 20 shell scripts and `py_compile` on the Python files.

Happy to adjust the model selection if some of these are known to do badly on Dutch, or to split the job scripts into a separate commit if that is easier to run in stages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ The Open ASR Leaderboard evaluates models on a diverse set of publicly available
   The [**ASR Longform benchmark**](https://huggingface.co/datasets/hf-audio/asr-leaderboard-longform) dataset includes earnings21 and earnings22. We also evaluate on [CORAAL](https://huggingface.co/datasets/bezzam/coraal), but it is stored as a separate dataset since it has multiple splits.
 
 * **Multilingual Benchmark:**
-  The [**ASR Multilingual benchmark**](https://huggingface.co/datasets/hf-audio/open-asr-leaderboard-multilingual-datasets) dataset includes fleurs, mcv and mls.
+  The [**ASR Multilingual benchmark**](https://huggingface.co/datasets/hf-audio/open-asr-leaderboard-multilingual-datasets) dataset includes fleurs, mcv and mls. For Hindi, data from [Voice Arena](https://huggingface.co/blog/open-asr-leaderboard-global-south) is used.
 
 
 * **Private datasets:** 
--- a/api/README.md
+++ b/api/README.md
@@ -31,7 +31,7 @@ bash api/run_api.sh
 
 Same idea, but using `api/run_api_ml.sh`, which evaluates on FLEURS, MCV (Mozilla Common Voice), and MLS (Multilingual LibriSpeech).
 - Set the relevant model in `MODEL_CONFIGS` of `api/run_api_ml.sh` with the model ID and the number max workers that the API allows, or pass `MODEL` via env var.
-- Set the relevant dataset/language pairs in `DATASET_CONFIGS` of `api/run_api_ml.sh`, or pass `"dataset:language"` pairs via `DATASETS` (see examples below).
+- Set the relevant dataset/language pairs in `DATASET_CONFIGS` of `api/run_api_ml.sh`, or narrow that list with `DATASETS` and `LANGUAGES` (see examples below).
 
 Smoke test (single model, single dataset/language):
 ```bash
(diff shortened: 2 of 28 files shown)


mrapaul:
Seed data for step 3, in case it saves a round of runs.

We already ran the two NeMo models on the three Dutch configs through this repo's `run_eval_ml.py`, scored with `score_results(multilingual=True, language="nl")`. Column order below matches `multilingual_es.csv` exactly, which is also what the `ml_nl` family in this PR emits:

```
model,RTFx,FLEURS WER,MCV WER,MLS WER
nvidia/canary-1b-v2,,5.27,5.78,9.64
nvidia/parakeet-tdt-0.6b-v3,,6.50,5.47,11.27
```

RTFx is deliberately left empty rather than filled in. Ours was measured on our own hardware, not your H200 flavour, and RTFx isn't portable across machines. Every row in the existing per-language CSVs carries either a real measurement or `-1`, and neither is honest for these two, so those cells should come from your run. Treat the WER columns as a cross-check against your numbers rather than as a submission.

Two things worth knowing before you compare:

Our Parakeet FLEURS-nl figure reproduces NVIDIA's published number to about 0.2pp. Separately, someone on #187 measured the same model at 6.20 on fleurs-nl against our 6.50, on different hardware and a different harness. A spread of that size between harnesses seems normal, but if your run lands closer to 6.2 than 6.5, that's likely why, not a wiring problem in this PR.

We also verified all three models saw a byte-identical reference set per config, by hashing the sorted references, with `n` and total duration matching.

No row for our own model here. It's closed-weight, so it can't run under HF Jobs, and a self-reported number has no business seeding a column.

mrapaul:
Rebased onto current `main` (past #207, the Granite turbo CTC merge). Still one clean diff, no conflicts.

Checked whether that merge left anything for this PR to cover: it adds `granite/run_eval_ctc.py` and `granite/submit_jobs.sh`, neither of which is a multilingual script, and the set of `*_ml.sh` files on `main` is unchanged. The new turbo CTC cards declare no languages, so I've not assumed Dutch for them. `granite/submit_jobs_nar_ml.sh` stays untouched for the same reason as before, its card lists en/fr/de/es/pt only.

Re-ran the checks after the rebase: `score_results(multilingual=True, language="nl", families=["ml_nl"])` still emits `model,RTFx,FLEURS WER,MCV WER,MLS WER`, which matches `multilingual_es.csv` byte for byte, a German-only results directory still scores exactly as before, and `bash -n` plus `py_compile` pass on everything touched.

Space side is prepared too, in https://huggingface.co/spaces/hf-audio/open_asr_leaderboard/discussions/71. That one is now two files: the `dutch` registry entry and `"dutch": "nl"` in `init.py`, plus the `EU_LANGUAGES` name and flag in `constants.py`, which I'd missed first time round. Safe to merge before `multilingual_nl.csv` exists either way, since the loader returns `None` for a missing file and the loop skips it.

MarktHart:
Might be relevant for you: https://huggingface.co/spaces/tvosch/Dutch-ASR-Leaderboard and it's repo https://github.com/tvosch/Dutch-ASR-leaderboard

ebezzam:
@mrapaul thanks for the PR here and on the leaderboard, I'll try to take a look soon!

ebezzam:
@mrapaul thanks again for the contribution! Dutch is ready to go 🇳🇱 

Below is a screenshot of the Dutch only results (draft) -> 12 of 29 models

Screenshot 2026-09-04 at 11 35 00

ebezzam:
results are live under the "Multilingual" tab!

PR #222, opened 2026-09-19
Add original LFM2.5-Audio-1.5B evaluation

kadirnar:
Local evaluation of the original `LiquidAI/LFM2.5-Audio-1.5B` on RTX 5070 Ti: 8 datasets, 74,443 utterances.

| Dataset | WER (%) |
| --- | ---: |
| AMI Cleaned | 13.83 |
| Earnings22 Cleaned AA (chunked) | 10.38 |
| GigaSpeech Cleaned | 9.27 |
| LibriSpeech test.clean | 1.58 |
| LibriSpeech test.other | 3.93 |
| SPGISpeech | 2.45 |
| Voice Arena Monsoon (en-IN) | 5.07 |
| VoxPopuli Cleaned AA | 2.94 |
| **Average** | **6.18** |

Aggregate local RTFx: **34.43**. H200 measurements are pending.


kadirnar:
I tested it with the Liquid-Audio repository. It has the same WER rate.

ebezzam:
thanks @kadirnar! I'll try to get in by the end of next week (maybe this one if time permits 😄)

liquid_audio/Dockerfile
ebezzam: could you instead make an HF space with the Dockerfile? Similar to the ones here: https://huggingface.co/collections/hf-audio/open-asr-leaderboard-eval-configurations

liquid_audio/requirements.txt
ebezzam: like the Dockerfile, we don't need this here and can be in the HF Space!

liquid_audio/submit_jobs.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
+# Dry run by default. RUN_LOCAL=1 evaluates locally; RUN_JOBS=1 submits HF Jobs.
+set -euo pipefail
+
+SPACE="${SPACE:-}"
```
ebezzam: you can set the HF space here

liquid_audio/submit_jobs.sh
```diff
@@ -0,0 +1,82 @@
...
+
+selected=0
+for entry in "${DATASETS[@]}"; do
+  read -r label config split dataset_path <<< "$entry"
+  if [[ -n "$ONLY_DATASETS" && " $ONLY_DATASETS " != *" $label "* ]]; then
+    continue
+  fi
+  selected=$((selected + 1))
+  [[ "$config" != - ]] || config=""
+  if [[ "$RUN_LOCAL" == 1 ]]; then
```
ebezzam: let's also inject the normalizer from the repo like [this](https://github.com/huggingface/open_asr_leaderboard/blob/6fec425b243962c0e7fd0b36b9b5841aef676c0a/transformers/submit_jobs_whisper.sh#L121)

ebezzam (commented):
Some comments to align with the other scripts/models

